PR #17, opened 2024-07-18
Code Climate settings

igoristic:
The `main` branch still needs to be enabled with Code Climate on the github side. I don't have the permissions to do that but it's possible to test it locally:

Test locally:
```bash
chmod +x test.sh
./test.sh
```

*We can also use this script with pre-commit hooks


IsabelLaurenceau (approved):
Approved! For future reference the rest of the team uses this template: https://cm.usa.gov/confluence/display/SRCH/Pull+Request+Template

test.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+if [ ! -f "/usr/local/bin/codeclimate" ]
+then
+	curl -L https://github.com/codeclimate/codeclimate/archive/master.tar.gz | tar xvz -C /home/$USER/
+	(cd /home/$USER/codeclimate-* && sudo make install)
```
selfdanielj, line 6: what about using just `$HOME/` here instead of `/home/$USER/`?  i was just testing this out locally and i don't have a `/home/$USER` but i do have a `$HOME` that points to `/users/dself` and I think `$HOME` will likely produce the same results on a linux based EC2 if thats where this is supposed to run otherwise.

search_gov_crawler/search_gov_spiders/spiders/domain_spider.py
```diff
@@ -1,9 +1,43 @@
...
+def valid_content_type(type_whitelist, content_type_header):
+        content_type_header = str(content_type_header)
+        for type_regex in type_whitelist:
+            if re.search(type_regex, content_type_header):
+                return True
+        return False
+
 starting_urls = os.path.join(
     os.path.dirname(sys.modules[__name__].__file__), "../utility_files/startingUrls.txt"
 )
```
selfdanielj, line 43: i know this piece didnt change but have you considered moving this and the reading of the file into the init method (or into a function that is called during init)?  This functionality seems like its only used if no domain (or url) is passed so there could be times when you are doing this work (and potentially causing an error) when you don't have to.

Also, do you ever use pathlib? for this stuff? I like the API a bit better than os, for instance you could replace reading and parsing the files that you have here with:
```python
starting_urls_file = Path(__file__).parent.parent / "utility_files" / "startingUrls.txt"
start_urls_list = starting_urls_file.resolve().read_text().splitlines()
```

.codeclimate.yml
```diff
@@ -0,0 +1,33 @@
...
+    config:
+      threshold: 300
+  method-lines:
+    enabled: false
+  method-count:
+    enabled: false
+  method-complexity:
+    enabled: false
+  file-lines:
+    enabled: false
```
selfdanielj, line 14: file-lines is disabled here and enabled on line 3... is that right? seems like it should only be one or the other?

test.sh
selfdanielj, line 1: Is this file going to persist in the repo?  If so, it might be more clear to name it `code_climate.sh` or `run_code_climate.sh` or something instead of `test.sh`.

.codeclimate.yml
```diff
@@ -0,0 +1,33 @@
...
+  fixme:
+    enabled: true
+  sonar-python:
+    enabled: true
+    exclude_patterns:
+    - "**/settings.py"
+  pep8:
+    enabled: true
+
+exclude_patterns:
```
selfdanielj, line 28: when i ran codeclimate i got INFO log messages like 
```
INFO: Invalid character encountered in file '/code/search_gov_crawler/__pycache__/__init__.cpython-38.pyc' at line 3 for encoding UTF-8. Please fix file content or configure the encoding to be used using property 'sonar.sourceEncoding'.
```

I dont think we need to analyze the pycache folders, right?  maybe - `"**/__pycache__/"` or maybe `__pycache__/` could be added to the `exclude_patterns` list?

search_gov_crawler/search_gov_spiders/spiders/domain_spider.py
igoristic: Yeah, this looks a lot cleaner, but I didn't want to change too much outside the scope of the ticket. We can create a cleanup/maintenance task and implement it everywhere.

test.sh
igoristic: Ah, this makes a lot more sense now why it didn't work on other machines. Strangely for me
`home/$USER == $HOME` 🤷 Probably a left over from some other scripts I was doing

IsabelLaurenceau (approved):

PR #24, opened 2024-08-14
Fixed unit tests and CircleCI

igoristic:


.circleci/config.yml
```diff
@@ -1,336 +1,32 @@
----
-version: 2
+version: 2.1
+
 jobs:
-  build_and_test:
+  build-and-test:
     docker:
-      - image: cimg/python:3.12
+      - image: cimg/python:3.9
```
selfdanielj, line 6: just curious, why did you decide to downgrade here?

igoristic: Some dependencies were failing to install, so I assumed it's because I changed the version number. But, I guess that wasn't it. Turns out some versions/packages were modified int the requirements.txt file manually, but it should be fixed now

selfdanielj: ok... i guess when we delete all but the scrapy libs we can upgrade again?

selfdanielj (approved):
Tests work locally for me.

PR #34, opened 2024-09-09
Code deploy build

igoristic:


cicd-scripts/app_install.sh
```diff
@@ -0,0 +1,60 @@
...
+    echo "Current Python version: $python_version"
+    echo "Installing Python 3.12..."
+    install_python
+fi
+
+
+
+# Creating python3.12 virtual env
+
+pip install virtualenv
```
selfdanielj, line 45: why do you install virtualenv here? doesnt look like you use it?

cicd-scripts/app_install.sh
```diff
@@ -0,0 +1,60 @@
...
+
+cd /home/ec2-user/python-flask-service
```
selfdanielj, line 47: why do you cd here?  isn't the code in `/home/ec2-user/spider`?

buildspec_spider.yml
selfdanielj, line 1: what is the purpose of this file?  it seems to do a lot of what the cicd scripts do.  Do we need both?

cicd-scripts/app_start.sh
```diff
@@ -0,0 +1,19 @@
...
+
+echo "Running searchgov-spider application..."
+# python /home/ec2-user/app/app.py
+
+pkill scrapy
+
+
+nohup scrapy crawl domain_spider &
+nohup scrapy crawl domain_spider_js &
+
```
selfdanielj, line 14: This should be starting scrapyd, scrapydweb, and logparser instead of scrapy directly right?

cicd-scripts/app_start.sh
```diff
@@ -0,0 +1,19 @@
...
+echo "\nCurrent running crawl jobs:"
+jobs
+
+echo "\noutput:"
+cat nohup.out
```
selfdanielj, line 19: do you think we will want to capture these logs in cloudwatch?  if so could stdout and stderr be redirected somewhere else?

cicd-scripts/app_install.sh
```diff
@@ -0,0 +1,60 @@
...
+
+echo "Creating python3.12 virtual environment..."
+python3.12 -m venv /home/ec2-user/app/venv
```
selfdanielj, line 50: did you install the venv in a different directory on purpose ("app" vs "spider")?  i dont think it matters just curious.

cicd-scripts/app_start.sh
selfdanielj: If so, im wondering if as a follow on we need some sort of logrotate.d or similar process to handle these.

appspec.yml
```diff
@@ -0,0 +1,34 @@
...
+    type:
+      - file
+  - object: /home/ec2-user/spider/cicd-scripts/app_stop.sh
+    owner: search
+    mode: 755
+    type:
+      - file
+hooks:
+  AfterInstall:
+    - location: spider/cicd-scripts/app_install.sh
```
selfdanielj, line 24: does this work without a fully resolved path? i don't know what the cwd is but probably not the project root

selfdanielj (commented):
A few questions below... if any change suggestions are valid, ill leave it up to you to decide if any should be included in this story or should be done in a follow on story.

igoristic:
@selfdanielj Thanks for all the feedback! Looks like I had some bad auto resolve conflicts with the multiple branches I force merged into 🤦‍♂️

cicd-scripts/app_install.sh
lsamuels-fearless, line 1: The version of Python is hard-coded here. is there a way for us to control this in a way that wouldn't require an app-side code change?

cicd-scripts/app_install.sh
```diff
@@ -1,51 +1,68 @@
 #!/bin/bash
 
+
+# A hack to get the environment running without ansible local env variables
+# This block of code will eventually be removed once ansible task is completed
+SPIDER_PYTHON_VERSION=3.12
+SPIDER_STAGING_URLS_API=https://staging.search.usa.gov/urls
```
selfdanielj, line 7: What about removing the environment name from this variable so we don't have to have two environment variables, SPIDER_STAGING_URLS_API and, i assume SPIDER_PROD_URLS_API when we can just have one SPIDER_URLS_API and set the value based on environment?

cicd-scripts/app_install.sh
```diff
@@ -1,51 +1,68 @@
...
                             tk-dev libgdbm-dev libc6-dev libbz2-dev \
                             zlib1g-dev openssl libffi-dev
 
-    # Download Python 3.12 source code
+    # Download Python source code
     cd /usr/src
-    sudo wget https://www.python.org/ftp/python/3.12.0/Python-3.12.0.tgz
-    sudo tar xzf Python-3.12.0.tgz
+    sudo wget https://www.python.org/ftp/python/${SPIDER_PYTHON_VERSION}.0/Python-${SPIDER_PYTHON_VERSION}.0.tgz
+    sudo tar xzf Python-${SPIDER_PYTHON_VERSION}.0.tgz
```
selfdanielj, line 38: does this mean we still have to make changes here for minor version updates?  Why not include the ".0" in the variable and the just use `python3` or `python` instead of `python 3.12` when upgrading pip below.

cicd-scripts/app_install.sh
```diff
@@ -1,51 +1,68 @@
...
+    echo "Python ${SPIDER_PYTHON_VERSION} is not installed. Installing Python ${SPIDER_PYTHON_VERSION}..."
     install_python
 fi
 
-# Install virtualenv using Python 3.12's pip
-sudo /usr/local/bin/python3.12 -m pip install --upgrade pip
-sudo /usr/local/bin/python3.12 -m pip install virtualenv
+# Install virtualenv using Python pip
+sudo /usr/local/bin/python${SPIDER_PYTHON_VERSION} -m pip install --upgrade pip
+sudo /usr/local/bin/python${SPIDER_PYTHON_VERSION} -m pip install virtualenv
```
selfdanielj, line 58: i don't see where we are using virtualenv.  Below you are using venv the built in module.  I don't think we need this installed.

selfdanielj (approved):
not sure if you were waiting on an approval but ok with me to merge whenever you are ready... you did a lot of great work here! 🎉

search_gov_crawler/search_gov_spiders/pipelines.py
```diff
@@ -4,47 +4,87 @@
...
+        self.file_number += 1
+        self.current_file = open(self.file_path, "w", encoding="utf-8")
+
+    def _post_urls(self):
+        """Send a POST request with the batch of URLs if any exist."""
+        if self.urls_batch:
+            try:
+                response = requests.post(self.api_url, json={"urls": self.urls_batch})
+                response.raise_for_status()
+                print(f"Successfully posted {len(self.urls_batch)} URLs to {self.api_url}.")
```
selfdanielj, line 76: I think you can probably access a logger here... can this be a `logger.info()` instead of a `print`?

search_gov_crawler/search_gov_spiders/pipelines.py
```diff
@@ -4,47 +4,87 @@
...
+            except requests.exceptions.RequestException as e:
+                print(f"Failed to send URLs to {self.api_url}: {e}")
```
selfdanielj, line 78: similar here, can this be `logger.error()`?

search_gov_crawler/search_gov_spiders/pipelines.py
```diff
@@ -4,47 +4,87 @@
...
+
+    def _rotate_file(self):
+        """Close current file, rename it, and open a new one for continued writing."""
+        self.current_file.close()
+        new_file_path = self.parent_file_path / f"output/all-links-{self.file_number}.csv"
+        os.rename(self.file_path, new_file_path)
+        self.file_number += 1
+        self.current_file = open(self.file_path, "w", encoding="utf-8")
+
+    def _post_urls(self):
```
selfdanielj, line 70: you talked about issues with staging being down and this causing lots of errrors in the logs... did you consider not catching the error here but instead raising it and failing the crawl?  Is there benefit it continuing if we can't write to this API?

search_gov_crawler/search_gov_spiders/pipelines.py
selfdanielj, line 1: is it possible to add any tests for this new functionality either through updating the full crawl tests or adding unit tests for the class?  Even just the private methods if you don't want to mock the post response.

cicd-scripts/app_install.sh
igoristic: This is based on semantic versioning, and some places in the code require only the major and a minor, eg: `3.12` If my variable also has the patch suffix `.0` then I would have to remove it where ever it's incompatible. So, basically it's easier to add the patch version rather than removing it. I didn't give it much thought though, because usually patch only has minor fixes/improvements and no breaking changes

cicd-scripts/app_install.sh
igoristic: Its aliased as `venv`. It might already be installed (usually it's part of python), but it's not guaranteed

selfdanielj: Did you alias it somewhere or is that built in functionality? venv is included in the standard python library.  

https://docs.python.org/3.12/library/venv.html
vs
https://pypi.org/project/virtualenv/

and from https://virtualenv.pypa.io/en/latest/:

> virtualenv is a tool to create isolated Python environments. Since Python 3.3, a subset of it has been integrated into the standard library under the [venv module](https://docs.python.org/3/library/venv.html). The venv module does not offer all features of this library, to name just a few more prominent.

search_gov_crawler/search_gov_spiders/pipelines.py
igoristic: Unit test's for bash scripts are not really conventional, since you'd have to mock most of the functionality and it's environment. Which makes it's really easy to give it a "happy path" where it passes the test, but would still be a false positive under a more realistic scenario

cicd-scripts/app_install.sh
igoristic: Ah you're right. I always thought they were the same thing and that we can install python without it and then add it via virtualenv

cicd-scripts/app_start.sh
```diff
@@ -0,0 +1,86 @@
...
+sudo pkill -f "scrapyd" 2>/dev/null
+
+# Check search-gov /urls endpoint
+echo "Checking search-gov /urls api..."
+if check_url "$SPIDER_URLS_API"; then
+    echo "The /urls api is up and running at endpoint: $SPIDER_URLS_API"
+else
+    echo "Error: /urls failed failed at endpoint: $SPIDER_URLS_API"
+    exit 1
+fi
```
selfdanielj, line 52: This means the spider app won't start if the URLs API is not available right?  Nothing in the scrapyd or scrapydweb processes themselves rely on the URLs API, its just the actually running jobs.  Should we just fail the job if the API is down? Why do we have to not even start these services?

selfdanielj:
Not sure if you've worked more on this but to fix the test_full_crawl errors you probably want to replace the mock_init function in test_full_crawl.py lines 96-102 with:
```python
        def mock_init(pipeline_cls, *_args, temp_dir=temp_dir, **_kwargs):
            pipeline_cls.api_url = None
            pipeline_cls.file_number = 1
            pipeline_cls.parent_file_path = temp_dir
            pipeline_cls.base_file_name = temp_dir / "output" / "all-links.csv"
            pipeline_cls.file_path = pipeline_cls.base_file_name
            pipeline_cls.current_file = open(pipeline_cls.file_path, "w", encoding="utf-8")
```			
or something similar.  I briefly looked at the test_url_file_size tests and noticed you are using the mocker fixture with pytests... did you mean to also add the pytest-mock plugin to the requirements file?  Besides that i didnt dig into it but seems like the mocks are not right.

cicd-scripts/app_install.sh
selfdanielj, line 1: the other day when i went to run things I noticed we needed these three libs installed for our app to work.  Can you add them to the apt-get?
lzma
liblzma-dev
libbz2-dev

selfdanielj, line 1: I think there was an issue with the `sudo pip install --upgrade` right? I was able to just remove the upgrade flag and have it work.

selfdanielj, line 1: should there be a line to install playwright?  i don't think thats ever been in here.

cicd-scripts/app_start.sh
igoristic: 🎣  Good catch! We are doing this check in the python code now.

cicd-scripts/app_install.sh
```diff
@@ -0,0 +1,132 @@
...
+echo "Creating python${SPIDER_PYTHON_VERSION} virtual environment..."
+sudo /usr/local/bin/python${SPIDER_PYTHON_VERSION} -m venv ./venv
+
+# Activate the virtual environment
+source ./venv/bin/activate
+
+# Install all spider dependencies
+echo "Installing dependencies..."
+sudo pip install --force-reinstall -r ./search_gov_crawler/requirements.txt
+sudo pip install pytest-playwright playwright -U
```
selfdanielj, line 97: why do we need these? should they be added to requirement.txt if they are needed?

cicd-scripts/app_install.sh
```diff
@@ -0,0 +1,132 @@
...
+playwright install
```
selfdanielj, line 98: i think you can probably just run `playwright install chromium` since that is what we are using in scrapy but this probably doesnt hurt.

cicd-scripts/app_install.sh
```diff
@@ -0,0 +1,132 @@
...
+else
+    echo "Python ${SPIDER_PYTHON_VERSION} is not installed. Installing Python ${SPIDER_PYTHON_VERSION}..."
+    install_python
+fi
+
+# Set PYTHONPATH env
+source ./cicd-scripts/helpers/update_pythonpath.sh
+
+# Use venv with Python 3.12
+sudo /usr/local/bin/python${SPIDER_PYTHON_VERSION} -m pip install --upgrade pip
```
selfdanielj, line 85: i thought the pip install usually happens inside the venv? do we need to upgrade the pip installed here?

cicd-scripts/helpers/run_without_ui.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+SPIDER_PYTHON_VERSION=3.12
+sudo bash -c "nohup /usr/local/bin/python${SPIDER_PYTHON_VERSION} ./search_gov_crawler/scrapy_scheduler.py > /var/log/scrapy.log 2>&1 &"
```
selfdanielj, line 3: does this work? usually you would want to point to the python in the venv?

IsabelLaurenceau (approved):

search_gov_crawler/search_gov_spiders/pipelines.py
```diff
@@ -2,46 +2,99 @@
...
+    def _batch_size(self):
+        """Calculate total size of the batched URLs."""
+        return sum(len(url.encode("utf-8")) for url in self.urls_batch)
+
+    def _file_size(self):
+        """Get the current file size."""
+        self.current_file.flush()  # Ensure the OS writes buffered data to disk
+        return self.file_path.stat().st_size
+
+    def _rotate_file(self):
```
IsabelLaurenceau, line 68: We need a way to identify the process /spider calling the file. This was done with my changes from https://github.com/GSA/searchgov-spider/pull/60 I recommend putting back my changes

IsabelLaurenceau (changes requested):
See my comment on pipelines.py

search_gov_crawler/requirements.txt
```diff
@@ -2,11 +2,13 @@ freezegun==1.5.1
...
 pytest-console-scripts==1.4.1
+pytest-mock==3.14.0
 
 python-json-logger==3.2.0
 scrapy==2.11.2
 scrapy-playwright==0.0.42
 scrapyd==1.5.0
 scrapyd-client==2.0.0
 scrapydweb @ git+https://github.com/GSA/searchgov-scrapydweb
+spidermon [monitoring] == 1.22.0
```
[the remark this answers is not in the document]
igoristic: Good catch 👍

cicd-scripts/helpers/run_without_ui.sh
igoristic: Good call!

I fixed this by activating the virtual environment

cicd-scripts/app_install.sh
igoristic: You're absolutely right

cicd-scripts/app_install.sh
igoristic: I installed it with the `--with-deps` flag

tests/search_gov_spiders/test_deduplicator_pipeline.py
selfdanielj, line 1: There was already some coverage of the de-deuplicator pipeline in `tests/search_gov_spiders/test_pipelines.py`... can you take a look and see if this can just replace that one?  I don't think we need tests in two places so maybe delete the other one if this fully replaces it?

igoristic: Ah, you're right I completely missed the one in `test_pipeline.py`. Just, thought we weren't 100% covered on codec-climate

PR #38, opened 2024-09-23
SRCH-5103 schedule tasks

selfdanielj:
## Summary
* This change sets up a **naive** schedule in scrapydweb.  Tasks are distributed evenly M-F except for a block held available for a maintenance window.  I wanted a time that we could do deployments or whatever else and not have to turn jobs off.
* The README file is mainly for documenting this starting point and can be removed once we move to prod.
* I added the scrutiny file here and made some updates to `crawl-sites.json` to include name and other fields we might use for scheduling.
* I also removed some of the un-scheduled crawls based on the scrutiny file and Jim's input.
* Let me know what you think about the schedule as a starting point or the init method.

Testing
* Follow the init instructions in the readme to schedule jobs.

### Checklist

#### Functionality Checks

#### Process Checks


IsabelLaurenceau (approved):
Overall looks good to me and works as expected. See below

nit: 
last command should be "sqlite3 <scrapydweb-data-dir>/database/timer_tasks.db < ./init_schedule.sql" not initialize_schedule.sql


future work/question: 
- In the init_schedule.sql we are setting all the tasks. Did you already create a script to make this or did you do it manually? If you did it manually I think a next step (next ticket no reason to hold this one up) would be to create a script that can do that automatically from the json since that's the format it will be anyway. If you did it via a script I think it should be added to the repo

selfdanielj:
> Overall looks good to me and works as expected. See below
> 
> nit: last command should be "sqlite3 /database/timer_tasks.db < ./init_schedule.sql" not initialize_schedule.sql
> 
> future work/question:
> 
>     * In the init_schedule.sql we are setting all the tasks. Did you already create a script to make this or did you do it manually? If you did it manually I think a next step (next ticket no reason to hold this one up) would be to create a script that can do that automatically from the json since that's the format it will be anyway. If you did it via a script I think it should be added to the repo

Thanks, i updated the readme.

I did it manually.  The three of us briefly chatted the other day in slack about if this should be automated or not.  I thought I remember us kind of agreeing that once we release it would be saved in the db and we could modify it from there but that doesn't mean we can't change our minds.  

I can make a script to do it (and I almost did) but the schedule info in scrutiny has things running all on Wednesday so its not really a good source.  Where should I store the schedule info?  Am i just hard coding it into the json? Or do I write something that automatically spreads them out over the week like i have? I couldn't really decide on the best way so I just did this SQL file manually.

IsabelLaurenceau:
No need to make the script now. When I saw the sql file I was more curious than anything for now. I think like you said this is a naive scheduler and as we run it and know more about how it works irl we can than put the effort into the script.

search_gov_crawler/search_gov_spiders/utility_files/crawl-sites.json
```diff
@@ -1,332 +1,324 @@
...
     {
+        "name": "VA Resources",
+        "job_id": "va-resources",
         "allowed_domains": "va.gov",
         "handle_javascript": false,
         "starting_urls": "https://www.va.gov/resources/"
     },
     {
+        "name": "DOD DANTES",
+        "job_id": "dod-dantes",
```
igoristic, line 319: nit: Do we really need this field? Can't we just convert the name to the `job_id` whenever we need, eg:
```python
name.lower().replace(" ", "-")
```

Unless we will need to use it as a unique identifier or as a reference later on

igoristic (approved):
Awesome job!! 🏆 

Yeah, I agree with @IsabelLaurenceau we should populate those rows from the crawl-sites.json file (which should also have the scheduled date/time intervals). Then the script can just be a python script that does bulk insert, eg:

```python
import sqlite3

conn = sqlite3.connect('database.db')
cursor = conn.cursor()

# Data to insert
data = [
    {
        "name": "DOD Army MWR",
        "job_id": "dod-army-mwr",
        "allowed_domains": "armymwr.com",
        "handle_javascript": false,
        "starting_urls": "https://www.armymwr.com/"
    },
    {
        "name": "DOD Defense Travel Management Office",
        "job_id": "dod-defense-travel-management-office",
        "allowed_domains": "travel.dod.mil",
        "handle_javascript": false,
        "starting_urls": "https://www.travel.dod.mil/"
    }...
]

cursor.executemany('INSERT INTO tasks (name, job_id, ...) VALUES (?, ?, ...)', data)
conn.commit()
conn.close()
```
We will need to do this every time we lose/delete our database (or if it gets corrupted)

selfdanielj:
I will create a follow on story to add a script similar to what is described here.

PR #44, opened 2024-10-10
SRCH-5547 Automate Schedule Initialization

selfdanielj:
## Summary
- Instead of running and maintaining the SQL statements we can just run a script now.  The full directions have been updated at `search_gov_crawler/search_gov_spiders/utility_files/README.md`
- Created init_schedule.py that takes a single argument, a json input file, that reads the input file, truncates tables, and inserts data into the task table.
  - Instead of hard-coding the schedule, created a class that generates the schedule.  Adding functionality that would wrap around the first set of schedule slots was going to complicate things and I don't think we need it so I just raise an error if someone tries to use it like that.
  - I am not sure of the best way to find the databases.  Please let me know especially if you have trouble with that part.  We can (and probably should) implement a DATA_PATH variable in prod but locally at least I'm just guessing.
- Removed job-id from crawl-sites.json based on [comment](https://github.com/GSA/searchgov-spider/pull/38#discussion_r1775521383) in first schedule PR
- Ran codecoverage and pylint locally and made small updates based on that feedback.
- Updated tests
  - New testing dependency `freezegun` that makes it easier to test current date/datetime values.  Added this to requirements, pyproject.toml and dependabot group.
  - added coverage where appropriate

### Testing
- Follow the directions in the readme linked above but if you have things running you can just do
```
$ cd search_gov_crawler/search_gov_spiders/utility_files
$ python init_schedule.py --input_file=./crawl-sites.json
```
- If this runs successfully you should be able to view, edit, and schedule the jobs in scrapydweb.

### Checklist

#### Functionality Checks

#### Process Checks


IsabelLaurenceau (approved):
Code looks good minus the conflicts that need to be addressed and works as expected! 

One of the things I looked at was what happens if I delete a few entries and run the code again. I noticed it replaced the entire table and when I looked through the code I saw that truncate_table() in init_scedule.py always deletes the entire schedule and re-uploads it. I think that makes sense for this requirement but I wonder what the future work of this is going to look like. Will we want more of an update table etc? (No need to speculate now since we should see how it's actually used but just a thought)

selfdanielj:
> of the things I looked at was what happens if I delete a few entries and run the code again. I noticed it replaced the entire table and when I looked through the code I saw that truncate_table() in init_scedule.py always deletes the entire schedule and re-uploads it. I think that makes sense for this requirement but I wonder what the future work of this is going to look like

Great point... l'll also keep this in mind going forward.

igoristic (approved):
Looks good 🏆

PR #48, opened 2024-10-30
SRCH-5588 Scrapy Scheduler

selfdanielj:
## Summary
- Allows for apscheduler-backed process to schedule and run `scrapy crawl` commands directly, without scrapyd or scrapydweb.  This approach does not add any new requirements nor does it break anything related to running scrapyd or scrapydweb.
- Updated readme and added a few basic tests.

### Testing
- Follow directions in main readme to start scheduler.  This will schedule all jobs according the the initial schedule specified in the [utilitiy files readme schedule](https://github.com/GSA/searchgov-spider/blob/main/search_gov_crawler/search_gov_spiders/utility_files/README.md#job-schedule-calendar). 
- You can now wait for the next scheduled run to ensure things work.

- If you don't want to wait you can try to schedule some test crawls by editing the `scrapy_scheduler.py` file to load crawl sites from the test sites file.  It will schedule the jobs 2, 4, and 6 minutes in the future. You can then watch the runs and monitor the output.
  - Replace `CRAWL_SITES_FILE` assignment on line 20 with `CRAWL_SITES_FILE = Path(__file__).parent.parent / "tests" / "search_gov_spiders" / "crawl-sites-test.json"`

  - Replace the entire `transform_crawl_sites` function with the code below. 
```python
def transform_crawl_sites(crawl_sites: list[dict]) -> list[dict]:
    """Transform crawl sites records into a format that can be used to"""

    transformed_crawl_sites = []
    schedule = SpiderSchedule()

    from datetime import datetime, UTC, timedelta

    current_datetime = datetime.now(tz=UTC)
    for idx, crawl_site in enumerate(crawl_sites):
        job_name = str(crawl_site["name"])
        schedule_slot = schedule.get_next_slot()

        next_run = current_datetime + timedelta(minutes=idx + 2)

        transformed_crawl_sites.append(
            {
                "func": run_scrapy_crawl,
                "id": job_name.lower().replace(" ", "-").replace("---", "-"),
                "name": job_name,
                "trigger": CronTrigger(
                    year="*",
                    month="*",
                    day="*",
                    week="*",
                    day_of_week=next_run.strftime("%a").lower(),
                    hour=next_run.hour,
                    minute=next_run.minute,
                    second=0,
                    timezone="UTC",
                    jitter=0,
                ),
                "args": [
                    "domain_spider" if not crawl_site["handle_javascript"] else "domain_spider_js",
                    crawl_site["allowed_domains"],
                    crawl_site["starting_urls"],
                ],
            }
        )

    return transformed_crawl_sites
```
### Checklist

#### Functionality Checks

#### Process Checks

### DIFF
--- a/search_gov_crawler/scrapy_scheduler.py
+++ b/search_gov_crawler/scrapy_scheduler.py
@@ -89,7 +89,7 @@ def start_scrapy_scheduler(input_file: Path) -> None:
     # Initalize scheduler
     scheduler = BlockingScheduler(
         jobstores={"memory": MemoryJobStore()},
-        executors={"default": ThreadPoolExecutor(max_workers=5)},
+        executors={"default": ThreadPoolExecutor(max_workers=10)},
         job_defaults={"coalesce": False, "max_instances": 1},
         timezone="UTC",
     )
--- a/search_gov_crawler/scrapy_scheduler.py
+++ b/search_gov_crawler/scrapy_scheduler.py
@@ -1,0 +1,106 @@
+import json
+import logging
+import subprocess
+import os
+from pathlib import Path
+
+from apscheduler.executors.pool import ThreadPoolExecutor
+from apscheduler.jobstores.memory import MemoryJobStore
+from apscheduler.schedulers.blocking import BlockingScheduler
+from apscheduler.triggers.cron import CronTrigger
+from pythonjsonlogger.json import JsonFormatter
+
+from search_gov_crawler.search_gov_spiders.extensions.json_logging import LOG_FMT
+from search_gov_crawler.search_gov_spiders.utility_files.init_schedule import SpiderSchedule
+
+logging.basicConfig(level=logging.INFO)
+log = logging.getLogger()
+log.handlers[0].setFormatter(JsonFormatter(fmt=LOG_FMT))
+
+CRAWL_SITES_FILE = Path(__file__).parent / "search_gov_spiders" / "utility_files" / "crawl-sites.json"
+
+
+def run_scrapy_crawl(spider: str, allowed_domains: str, start_urls: str) -> None:
+    """Runs `scrapy crawl` command as a subprocess given the allowed arguments"""
+
+    scrapy_env = os.environ.copy()
+    scrapy_env["PYTHONPATH"] = str(Path(__file__).parent.parent)
+
+    subprocess.run(
+        f"scrapy crawl {spider} -a allowed_domains={allowed_domains} -a start_urls={start_urls}",
+        check=True,
+        cwd=Path(__file__).parent,
+        env=scrapy_env,
+        executable="/bin/bash",
+        shell=True,
+    )
+    log.info(
+        "Successfully completed scrapy crawl with args spider=%s, allowed_domains=%s, start_urls=%s",
+        spider,
+        allowed_domains,
+        start_urls,
+    )
+
+
+def transform_crawl_sites(crawl_sites: list[dict]) -> list[dict]:
+    """Transform crawl sites records into a format that can be used to"""
+
+    transformed_crawl_sites = []
+    schedule = SpiderSchedule()
+
+    for crawl_site in crawl_sites:
+        job_name = str(crawl_site["name"])
+        schedule_slot = schedule.get_next_slot()
+        transformed_crawl_sites.append(
+            {
+                "func": run_scrapy_crawl,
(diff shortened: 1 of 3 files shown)


search_gov_crawler/scrapy_scheduler.py
```diff
@@ -0,0 +1,106 @@
...
+
+    # Load and transform crawl sites
+    crawl_sites = json.loads(input_file.read_text(encoding="UTF-8"))
+    apscheduler_jobs = transform_crawl_sites(crawl_sites)
+
+    # Initalize scheduler
+    scheduler = BlockingScheduler(
+        jobstores={"memory": MemoryJobStore()},
+        executors={"default": ThreadPoolExecutor(max_workers=5)},
+        job_defaults={"coalesce": False, "max_instances": 1},
```
igoristic, line 93: Should we maybe max these out to the instance's available resources. Not sure how it works, but usually if you don't provide the params (max_workers and max_instances) it will use the system's max

igoristic (approved):
Awesome job @selfdanielj !! ⭐ 

Approved, however, I think we should also remove all scrapyd/scrapydweb modules from requirements and all its code. And, add it all back once we decide to use them again. Otherwise, we'll have a lot of dead code

search_gov_crawler/scrapy_scheduler.py
selfdanielj: For max workers... looks like it defaults (for apscheduler at least) to 10 https://apscheduler.readthedocs.io/en/stable/modules/executors/pool.html?highlight=threadpoolexecutor#apscheduler.executors.pool.ThreadPoolExecutor. Sure we could increase should it be a environment variable we could set upstream in ansible or something?  
```python
executors={"default": ThreadPoolExecutor(max_workers=os.environ.get("AP_SCHEDULER_MAX_WORKERS", 10))}
```
default could be 10 or 20 or whatever but i prefer an explicit call out here even if we go with 10.

for max_instances, i couldn't find it documented but in the BaseScheduler class it appears the default is 1:
https://github.com/agronholm/apscheduler/blob/2a2210fc71683547f8b01c7d4552815c3b67c1fe/apscheduler/schedulers/base.py#L714

This is how many instances of the crawl job can run at once.  This seems like it should be left at 1, right?  Do we want to be running a crawl of example.com and then start another example.com crawl before that first one finishes?

The other day I did some manual testing of what happens when you hit these numbers and apscheduler just logs a warning with a reason and skips the run.

selfdanielj:
> Awesome job @selfdanielj !! ⭐
> 
> Approved, however, I think we should also remove all scrapyd/scrapydweb modules from requirements and all its code. And, add it all back once we decide to use them again. Otherwise, we'll have a lot of dead code

I think that's well intentioned but I wonder if its worth it to do all that.  I thought this was just supposed to be "back up" to the real solution of MFAing scrapydweb.  Isn't the next thing you are working on adding in login.gov?  won't you need all the scrapyd/scrapydweb code in the repo to do that? @IsabelLaurenceau what do you think?

search_gov_crawler/scrapy_scheduler.py
igoristic: Ah, okay makes sense. Yeah, I think 10 is a good start then. Usually ec2 instances have 2 threads per core so we can assume: `max_workers=instance_cpus*2`

IsabelLaurenceau:
> > Awesome job @selfdanielj !! ⭐
> > Approved, however, I think we should also remove all scrapyd/scrapydweb modules from requirements and all its code. And, add it all back once we decide to use them again. Otherwise, we'll have a lot of dead code
> 
> I think that's well intentioned but I wonder if its worth it to do all that. I thought this was just supposed to be "back up" to the real solution of MFAing scrapydweb. Isn't the next thing you are working on adding in login.gov? won't you need all the scrapyd/scrapydweb code in the repo to do that? @IsabelLaurenceau what do you think?

Yeah I agree with @selfdanielj I don't think we should remove scrapyd or scrapydweb if this is only meant to be a backup. If we do end up going with this and for some reason MFA and UI integration is delayed for a foreseeable future than I think we can and should discuss removing them to not have dead code but I think for now they should stay.

IsabelLaurenceau (approved):

search_gov_crawler/scrapy_scheduler.py
selfdanielj: i made the update to hardcode to 10... if we want to adjust or parameterize later we can do that.

PR #56, opened 2024-11-25
SRCH-5714 add ability to do a single benchmark run

selfdanielj:
## Summary
Created this draft PR to help facilitate bench marking... what about something like this? gives us some flexibility?
```

python search_gov_crawler/benchmark.py --starting_urls="https://quotes.toscrape.com" --allowed_domains="quotes.toscrape.com"
```
This will add a job using a similar process as would with the no-UI functionality (and also scrapydweb I think) but we can control what we pass in.

If we need more configurable options we can add more environment variables or add a config file.

### Checklist

#### Functionality Checks

#### Process Checks

### DIFF
--- a/search_gov_crawler/benchmark.py
+++ b/search_gov_crawler/benchmark.py
@@ -1,0 +1,68 @@
+import argparse
+import logging
+import time
+import os
+from datetime import datetime, UTC, timedelta
+
+from apscheduler.executors.pool import ThreadPoolExecutor
+from apscheduler.jobstores.memory import MemoryJobStore
+from apscheduler.schedulers.background import BackgroundScheduler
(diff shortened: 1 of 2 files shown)


selfdanielj:
> Code change in the scheduler makes sense to me. I'm not sure if it was just for an example but think we would want to use actual federal domains and not quotes to scrape.

For sure... you can pass in any domain/starting url and also enable javascript
`python search_gov_crawler/scrapy_scheduler.py --starting_urls="https://www.fac.gov/" --allowed_domains="fac.gov"`
`python search_gov_crawler/scrapy_scheduler.py --starting_urls="https://www.js.example/" --allowed_domains="js.example" --handle_js`

PR #59, opened 2024-12-10
Srch 5744 benchmark findings

selfdanielj:
## Summary
* Made updates and re-organized main `settings.py` so that:
  * Settings adjusting similar high-level behavior are next to each other in the file
  * Commented out a few settings that did not apply because the main flag for that setting is disabled.
  * Disabled the [telnet console](https://docs.scrapy.org/en/latest/topics/telnetconsole.html) since we are not using it, I don't see us starting to use it.  Disabling it likely frees up some memory/CPU that can be used elsewhere.
  * Disabled the close spider extension since I think we want to make sure we crawl every link available and don't want to stop crawling prematurely.
  * Enforced a one second delay between requests and a single concurrent connection.
  * Moved all playwright related settings to the `domain_spider_js` spider to prevent any usage of memory or CPU loading playwrght during non-JS runs as was observed in the benchmark testing.
* Added a more descriptive logging representation of spiders (and crawlers) so that we can identify log messages that belong to different spiders in the same output stream.  Previously you might have seen only `{"spider": "domain_spider"}` or `{"crawler": "search_gov_spiders"}` at the end of log messages but now you should see:

```
{
    "spider": {
        "name": "domain_spider",
        "allowed_domains": "example.com",
        "starting_urls": "https://www.example.com"
    }
}
OR
{
    "crawler": {
        "name": "search_gov_spiders"
    }
}
```

### How to Test
Run a spider e.g. `scrapy crawl domain_spider -a allowed_domains=quotes.toscrape.com -a start_urls=https://quotes.toscrape.com/` and verify the following:
* Speed of requests, as reported in the logs, is limited to a maximum of 1 page per second (60/min)
* No log messages printed about telnet starting
* While the spider is running, in another terminal run `ps aux | grep playwright`.  There should be no results.  Compare this to main to see the change.
* Ensure output correct and where expected
* Verify that logging messages specific to a spider or crawler contains a detailed spider or crawler object at the end of the log message as shown above.

Run a js spider e.g.`scrapy crawl domain_spider_js -a allowed_domains=quotes.toscrape.com -a start_urls=https://quotes.toscrape.com/js/`
* Ensure output correct and where expected
* Similar logging as above.

### Checklist

#### Functionality Checks

#### Process Checks

### DIFF
--- a/search_gov_crawler/search_gov_spiders/extensions/json_logging.py
+++ b/search_gov_crawler/search_gov_spiders/extensions/json_logging.py
@@ -11,13 +11,17 @@ from scrapy.signals import spider_opened
 LOG_FMT = "%(asctime)%(name)%(levelname)%(message)"
 
 
-def search_gov_default(obj) -> str | None:
-    """Function to help serialize scrapy objects"""
+def search_gov_default(obj) -> dict | None:
+    """Function to help serialize scrapy objects in logs"""
     if isinstance(obj, Spider):
-        return obj.name
+        return {
+            "name": obj.name,
+            "allowed_domains": getattr(obj, "allowed_domains", None),
+            "start_urls": obj.start_urls,
+        }
 
     if isinstance(obj, Crawler):
-        return str(obj.settings.get("BOT_NAME", "Unknown"))
+        return {"name": str(obj.settings.get("BOT_NAME", "Unknown"))}
 
     return None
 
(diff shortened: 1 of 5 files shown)


IsabelLaurenceau (approved):
Logging and functionality of new settings looks good! 

Screenshot 2024-12-10 at 3 00 51 PM
Screenshot 2024-12-10 at 3 12 17 PM

igoristic (approved):
Looks good 🏅

PR #60, opened 2024-12-10
SRCH-5748-Update-Spider-FEEDS-Setting

IsabelLaurenceau:
## Summary
We needed to disable to feed exporter (to do this just remove it from the settings.py) so that when the spiders run in parallel the extra file is not written to superfluously. 

Added file lock for writing to the all-links Item pipeline. 

Each spider will write to it's own all-links csv (which means each job has it's own file). Many spiders will not hit the 3.9MB limit but this will allow future work to immediately send the results (completed csv file) as the spider/job closes without waiting for the file to fill up for a new spider to populate it.  

#### Testing 
When you run a benchmark test make sure no file is being written to search_gov_crawler/scrapy_urls/domain_spider or search_gov_crawler/scrapy_urls/domain_spider_js. Only the all-links csvs should be created now. 

### Checklist

#### Functionality Checks

#### Process Checks

### DIFF
--- a/search_gov_crawler/search_gov_spiders/pipelines.py
+++ b/search_gov_crawler/search_gov_spiders/pipelines.py
@@ -4,10 +4,9 @@ See: https://docs.scrapy.org/en/latest/topics/item-pipeline.html
 """
 
 import os
+import threading
 from pathlib import Path
-
 from scrapy.exceptions import DropItem
-
 
 class SearchGovSpidersPipeline:
     """
@@ -23,27 +22,27 @@ class SearchGovSpidersPipeline:
         self.short_file = open(self.base_path_name, "w", encoding="utf-8")
         self.max_file_size = 3900
         self.paginate = True
+        self.lock = threading.Lock()
 
     def process_item(self, item, _spider):
         """Checks that the file is not at max size.
         Adds it to the file if less, or creates a new file if too large."""
+        with self.lock: 
+            line = item["url"]
+            self.current_file_size += 1
+            next_file_size = self.current_file_size + len(line)
+            if self.paginate and next_file_size > self.max_file_size:
+                self.short_file.close()
+                new_name = str(self.parent_file_path / f"output/all-links{self.file_number}.csv")
+                os.rename(self.base_path_name, new_name)
+                self.file_number = self.file_number + 1
+                self.short_file = open(self.base_path_name, "w", encoding="utf-8")
+                self.current_file_size = 0
 
-        line = item["url"]
-        self.current_file_size += 1
-        next_file_size = self.current_file_size + len(line)
-        if self.paginate and next_file_size > self.max_file_size:
-            self.short_file.close()
-            new_name = str(self.parent_file_path / f"output/all-links{self.file_number}.csv")
-            os.rename(self.base_path_name, new_name)
-            self.file_number = self.file_number + 1
-            self.short_file = open(self.base_path_name, "w", encoding="utf-8")
-            self.current_file_size = 0
-
-        self.short_file.write(line)
-        self.short_file.write("\n")
-        self.current_file_size = self.current_file_size + len(line)
-
-        return item
+            self.short_file.write(line)
+            self.short_file.write("\n")
+            self.current_file_size = self.current_file_size + len(line)
+            return item
 
 
 class DeDeuplicatorPipeline:
--- a/search_gov_crawler/search_gov_spiders/pipelines.py
+++ b/search_gov_crawler/search_gov_spiders/pipelines.py
@@ -4,7 +4,6 @@ See: https://docs.scrapy.org/en/latest/topics/item-pipeline.html
 """
 
 import os
-import threading
 from pathlib import Path
 from scrapy.exceptions import DropItem
 import fcntl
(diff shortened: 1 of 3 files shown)


selfdanielj:
I confirmed that the default FEEDS output is no longer being produced.

I ran into a potential issue while investigating the custom pipeline.  My initial concern was that the threading.lock() solution would slow things down a lot, that doesn't seem to be the case and I would guess it is probably working to block other threads from writing to the file however I think there is still a collision going on, probably with the names.  I haven't exactly worked it out yet but I would expect if I run two spiders in parallel the output files should contain the same number of links as they would if i ran the two spiders individually.  Is that right?  If so, thats not what I'm seeing.  Below i have some steps to reproduce.  

I was expecting the total link count in the parallel test to be 498... or maybe something else but certainly not LESS than the first test.  Instead it looks like at least all-links1.csv is being overwritten somehow?  Anyway seems to me like something weird is going on.  What do you think?

```
cd search_gov_crawler
rm output/*.csv
scrapy crawl domain_spider -a allowed_domains="quotes.toscrape.com" -a start_urls="https://quotes.toscrape.com"
find ./output -name "*.csv" | xargs wc -l
      55 ./output/all-links.csv
      81 ./output/all-links1.csv
      82 ./output/all-links3.csv
      79 ./output/all-links2.csv
      81 ./output/all-links4.csv
     378 total
ls -lh output/*.csv
-rw-r--r--  1 dself  staff   2.5K Dec 11 13:56 output/all-links.csv
-rw-r--r--  1 dself  staff   3.8K Dec 11 13:56 output/all-links1.csv
-rw-r--r--  1 dself  staff   3.8K Dec 11 13:56 output/all-links2.csv
-rw-r--r--  1 dself  staff   3.8K Dec 11 13:56 output/all-links3.csv
-rw-r--r--  1 dself  staff   3.8K Dec 11 13:56 output/all-links4.csv

rm output/*.csv
scrapy crawl domain_spider -a allowed_domains="quotes.toscrape.com/tag/" -a start_urls="https://quotes.toscrape.com"
      40 ./output/all-links.csv
      80 ./output/all-links1.csv
     120 total

ls -lh output/*.csv	 
-rw-r--r--  1 dself  staff   1.9K Dec 11 13:57 output/all-links.csv
-rw-r--r--  1 dself  staff   3.8K Dec 11 13:57 output/all-links1.csv

EDIT tests/search_gov_spiders/crawl-sites-test.json to remove jobs "Quotes 2" and "Quotes 3" so that we are running the two tests above in parallel.

rm output/*.csv
python benchmark.py -f ../tests/search_gov_spiders/crawl-sites-test.json
find ./output -name "*.csv" | xargs wc -l
      55 ./output/all-links.csv
      41 ./output/all-links1.csv
      82 ./output/all-links3.csv
      80 ./output/all-links2.csv
      81 ./output/all-links4.csv
     339 total

ls -lh output/*.csv
-rw-r--r--  1 dself  staff   2.5K Dec 11 13:53 output/all-links.csv
-rw-r--r--  1 dself  staff   1.9K Dec 11 13:53 output/all-links1.csv
-rw-r--r--  1 dself  staff   3.8K Dec 11 13:53 output/all-links2.csv
-rw-r--r--  1 dself  staff   3.8K Dec 11 13:53 output/all-links3.csv
-rw-r--r--  1 dself  staff   3.8K Dec 11 13:53 output/all-links4.csv
```

igoristic (approved):
Looks good ⭐

search_gov_crawler/search_gov_spiders/pipelines.py
```diff
@@ -5,9 +5,8 @@
 
 import os
 from pathlib import Path
-
 from scrapy.exceptions import DropItem
-
+import fcntl
```
selfdanielj, line 9: if you are naming the files uniquely, do you still need the fnctl lock?

tests/search_gov_spiders/test_full_crawl.py
```diff
@@ -101,6 +102,7 @@ def mock_init(pipeline_cls, *_args, temp_dir=temp_dir, **_kwargs):
             pipeline_cls.short_file = open(pipeline_cls.base_path_name, "w", encoding="utf-8")
             pipeline_cls.max_file_size = 3900
             pipeline_cls.paginate = True
+            pipeline_cls.lock = threading.Lock()
```
selfdanielj, line 105: i don't think we want this in the test anymore, right?

search_gov_crawler/search_gov_spiders/pipelines.py
```diff
@@ -19,30 +18,31 @@ def __init__(self, *_args, **_kwargs):
         self.current_file_size = 0
         self.file_number = 1
         self.parent_file_path = Path(__file__).parent.parent.resolve()
-        self.base_path_name = str(self.parent_file_path / "output/all-links.csv")
-        self.short_file = open(self.base_path_name, "w", encoding="utf-8")
-        self.max_file_size = 3900
+        self.base_path_name = str(self.parent_file_path / f"output/all-links-p{os.getpid()}.csv")
```
selfdanielj, line 21: i guess pid+YYYYMMDD might be more unique since they are technically reused but pid by iteself will probably do.  I guess our hope is not to really use the file writing option too often.

IsabelLaurenceau: I'm not sure I understand what you mean by they are technically reused?

search_gov_crawler/search_gov_spiders/pipelines.py
IsabelLaurenceau: We are still using multiple threads per spider.

search_gov_crawler/search_gov_spiders/pipelines.py
IsabelLaurenceau: Ah you mean not in the same run? I didn't assume these were being saved week to week here. All the ones in drive are overwritten every week

tests/search_gov_spiders/test_full_crawl.py
selfdanielj: isn't it a totally different library?  there isn't even a `lock` instance variable in the actual code.

search_gov_crawler/search_gov_spiders/pipelines.py
selfdanielj: Are we? I really don't know... haha.  I removed it and still go the same total count while running two spiders.  Fine to leave in.

IsabelLaurenceau: Yeah, in the item pipeline specifically. -> https://docs.scrapy.org/en/latest/topics/settings.html#std:setting-CONCURRENT_REQUESTS_PER_IP:~:text=the%20startproject%20command.-,CONCURRENT_ITEMS,-%C2%B6

IsabelLaurenceau: I could change this back to the with lock that it was before if we don't want to import fnctl

selfdanielj (approved):
Looks good!

PR #66, opened 2024-12-19
Srch 5764 sandbox

IsabelLaurenceau:
## Summary
Hello! 

I'm including some notes and potential gotcha's of things I ran into while fiddling with it  this week. Once you are in your sandbox space in the cli you should run "cf push <APP-NAME>" from the searchgov-spider directory.  (APP-NAME is whatever you want. I match mine to the manifest - spider-sandbox ). This will automatically create the app and start running the commands in startcommand.sh The new files in this directory are 
- .profile
    - [Allows us to update our python path.](https://docs.cloudfoundry.org/devguide/deploy-apps/deploy-app.html#:~:text=.-profile-script%20that%20contains) 
- manifest.yml
    - Needed to specify the buildpack, no-route and health check 
- requirements.txt
- runtime.txt
    - Used to change the version of python in the droplet. Default used is 3.10[ this will fail](https://github.com/open-webui/open-webui/discussions/4951#:~:text=In%20Python%20%3E%3D%203.11%20UTC%20is%20implemented%20in%20the%20datetime%20function%2C%20but%20in%20Python%203.10%20it%20is%20implemented%20in%20the%20timezone%20function.).  I have set it to use the version 3.12.x
- startcommand.sh
    - Tell the app what to do. If you want to run different commands you can do it from here or explained below. 
    - If this is not working make sure you add permissions to execute: chmod +x startcommand.sh

After you run cf push the first time you will most likely get either an error from requirements.txt complaining about "scrapydweb @ git+https://github.com/GSA/searchgov-scrapydweb" or the error " Connection was refused by other side: 111: Connection refused." or  when trying to run a crawl.  You need to update your[ egress traffic rules. ](https://cloud.gov/docs/management/space-egress/) You can do that by running "cf bind-security-group public_networks_egress sandbox-gsa --space <SPACE> " i.e my <SPACE> is my name. 

I mentioned you can use startcommand.sh to run different commands. i.e  running 
- python benchmark.py -f ./test_parallel_domain.json 
- python benchmark.py -d "quotes.toscrape.com" -u "https://quotes.toscrape.com"
- python scrapy crawl domain_spider -a allowed_domains="quotes.toscrape.com/tag/" -a start_urls="https://quotes.toscrape.com"

You can also use "python -m http.server" and in another terminal use "cf ssh <APP-NAME> to pass in the scrapy commands you want to test there. (Or do whatever else you want to see while the app is consistently running). 

If you are using the startcommand.sh to just run one command (benchmark, one spider crawl or scheduler that actually finishes) the app will end. I suggest you run cf stop <APP-NAME> as when the app closes [cloud.gov](http://cloud.gov/) will just restart it and run it all again.
### Checklist

#### Functionality Checks

#### Process Checks


requirements.txt
selfdanielj, line 1: I was able to deploy after replacing the contents of this file with what I have below.  This way we don't have to maintain two requirements.txt files.
```txt
#  support cloud.gov deployment without duplicating requirements file
-r search_gov_crawler/requirements.txt
```

startcommand.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+echo "BEGIN START COMMAND"
+cd search_gov_crawler
+python benchmark.py -f ./test_domains.json
```
selfdanielj, line 4: test_domains.json doesn't exist.  Seems like you could repoint to the file in `tests/search_gov_spiders/crawl-sites-test.json`

IsabelLaurenceau: You are right, we should have it point to that. I had made test_domains.json as a smaller file that would finish for testing locally

selfdanielj (approved):
Nice job! Deployment and run works for me as designed.

igoristic (approved):
Looks good 🏅

PR #67, opened 2024-12-27
SRCH-5745 Scheduler Update

selfdanielj:
## Summary
- Since the "naive" scheduler implemented in `search_gov_crawler/search_gov_spiders/utility_files/init_schedule.py` is not optimal for a polite spider, we need to design something else.  Additionally, I wanted it to be more flexible.  To that end I added a "schedule" field in the crawl-sites.json file that holds a cron expression that can be fed, as-is, into the scheduler.  A record will now look like:
```json
    {
        "name": "DOD Army MWR",
        "allow_query_string": false,
        "allowed_domains": "armymwr.com",
        "handle_javascript": false,
        "schedule": "00 14 * * MON",
        "starting_urls": "https://www.armymwr.com/"
    }
```
With each row defining when and how often that domain will run -- in this case every Monday at 2pm UTC.  We can make changes and redeploy as needed.  I added some dataclasses to help organize and validate this structure.
- I updated scrapy_scheduler.py and benchmark.py to use this functionality.
- I updated tests and also added tests for benchmark since they were missing.
- I updated the main README to describe this usage.
- I decided to just leave the old naive scheduler code in place for now.
- I also updated the plist import even though i don't really see us using it but I've been keeping it up to date all this time so why stop now?

### Initial Schedule
If you all have any feedback on this part especially, please let me know!  I looked at a few options and decided to start us off with a limit of 5 concurrent spider processes.  Sure we can do 50 but why? Given the lengths of some of the longer running jobs, we are just going to make things more complicated to debug if we have an error during the few minutes where 50 processes are running.  We can always adjust but also given what we don't know about the /urls endpoint I thought it best to be conservative.   Given this schedule we can finish everything (except NOAA Coastwatch which runs for at least a week) in 13ish hours.

We easily can change the limit with an environment variable and we can change the schedule in the crawl-sites.json file.

Because the JS sites take up so many resources, i put them both first and gave them ample time to finish before the nest job starts.  Then I scheduled the five longest running jobs and then all the rest.  As slots are opened up by jobs finishing, another job should start. I have estimates for all job run times [here](https://docs.google.com/spreadsheets/d/1PPx78a10I22EqeXNztqVIgzgK8T9WENcHhvlzYfNrSQ/edit?gid=0#gid=0).  I left an hour gap to ensure all the long running jobs have started (probably could have been five minutes) so that all the rest get queued properly. 

I just guessed on the start time (0600).  All times are in UTC.

0600 - Navy All Hands Magazine (JS site)
0730 - USDA NFC Help Documentation (JS Site)
0830 - Long running domains (DOD Army MWR, NOAA CoastWatch, NOAA CoastWatch East Coast Node, VA Benefits.va.gov, and VA Research & Development)
0930 - Rest of domains

### Testing
To test the scrapy_scheduler process, you can edit the craw-sites.json file and change one of the domains so that the schedule is a few minutes into the future -- if its 10:05 AM EST you can replace the schedule value with "07 15 * * *" for instance so that the particular domain will run at every day at 15:07 UTC or 10:07 EST and then run `python search_gov_crawler/scrapy_schedule.py`.  It will add all the domains to the scheduler and the one you changed should run when you expect.

You can also test the functionality with the benchmark process.  The easiest way is to just point the benchmark process to the crawl-sites-test.json file. Those domains are setup to run every 15 minutes, i.e. "0,15,30,45 * * * *".  To do this run `python search_gov_crawler/benchmark.py -f tests/search_gov_spiders/crawl-sites-test.json`

You should also verify the crawl-sites.json file to make sure the cron expressions match what I have described.

### Checklist

#### Functionality Checks

#### Process Checks

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,9 @@ The home for the spider that supports search.gov.
   * [Running Against A Specific Domain](#running-against-a-specific-domain)
 * [Setup and Use](#setup-and-use)
   * [Option 1: command-line](#option-1-straight-from-command-line)
-  * [Option 2: server](#option-2-deploying-on-server-scrapyd)
+  * [Option 2: benchmark](#option-2-benchmark-command-line)
+  * [Option 3: custom scheduler](#option-3-custom-scheduler)
+  * [Option 4: scrapyd](#option-2-deploying-on-server-scrapyd)
 * [Adding New Spiders](#adding-new-spiders)
 * [Running Scrapydweb UI](#running-scrapydweb-ui)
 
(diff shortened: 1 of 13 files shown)


igoristic (approved):
Looks good 👍

IsabelLaurenceau (approved):

PR #80, opened 2025-01-27
SRCH-5816 Pre-Prod Adjustments

selfdanielj:
## Summary
As discussed this change takes care of a few known issues:
* Removes `drive.hhs.gov` from the crawl list
* Adds a 24 hour item timeout.

Also, in staging I noticed that jobs were not firing as expected... I realized that the settings used in `benchmark` to operate the schedule like a queue -- meaning five running slots and the rest of them wait for one to be finished before running -- were not applied to the `scrapy_scheduler` code.  Up until now we were not having successful deploy and thus not running scrapes from the `scrapy_scheduler` code, just from `benchmark` I made that change and deployed a test branch to confirm correct behavior.

### Testing
* If you want to test the timeout the easiest way is to change the timeout value in the settings file to 1 and manually run a js domain.  It should exit prior to capturing any URLs.  Given that this is scrapy core functionality I don't think we need to worry about testing it too much.
* You can verify the scheduler works by adjusting the times in the crawl sites file to something closer to your current time and running... you can also check in dev starting at 2025-01-27 1900UTC to see the results of the scheduler I kicked off with these changes but with a change to the schedule to start at 1900.  You should see multiple jobs ending and starting.  Compare this to dev starting at 2025-01-27 0600 UTC which is the current main code running, you can see that after 930 more jobs do not start when they should.

### Checklist

#### Functionality Checks

#### Process Checks

### DIFF
--- a/search_gov_crawler/benchmark.py
+++ b/search_gov_crawler/benchmark.py
@@ -57,7 +57,7 @@ def init_scheduler() -> BackgroundScheduler:
     return BackgroundScheduler(
         jobstores={"memory": MemoryJobStore()},
         executors={"default": ThreadPoolExecutor(max_workers)},
-        job_defaults={"coalesce": False, "max_instances": 1},
+        job_defaults={"coalesce": True, "max_instances": 1, "misfire_grace_time": True},
         timezone="UTC",
     )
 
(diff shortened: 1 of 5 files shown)


igoristic (approved):
Looks good 👍

PR #81, opened 2025-01-30
Update scrapy_scheduler.py

IsabelLaurenceau:
## Summary
Added the option to use an environment variable for the CRAWL_SITES_FILE (provided the hardcoded as a default value if needed). 
This will allow us to run different schedules easily on dev vs stage vs prod

Though a default is provided. Suggested to wait until ansible playbooks have been updated with env_var
### Checklist

#### Functionality Checks

#### Process Checks


tests/search_gov_spiders/test_crawl_sites.py
selfdanielj, line 1: i think this should still point to the production file since its purpose is to validate the file, making sure that the json will actually start a scheduler.  Or maybe this should actually do both files if they are both going to be used.  You could parameterize this test case to accomplish that.

selfdanielj (commented):
Looks good except for my one question about the tests.

Is the idea that `crawl-sites-sample.json` will be used in dev and staging?  If so did you consider making it smaller?  I would think at least one JS and one non-JS would be good.

tests/search_gov_spiders/test_crawl_sites.py
IsabelLaurenceau: The reason we want to separate the production file out is because we are worried about resources on dev and staging so I don't think it should point to that file. I didn't consider changing the sample file since it's what we already use and it doesn't seem to be a resource issue.

selfdanielj: i still think the `test_crawl_sites_file_is_valid` test in `test_crawl_sites.py` file should test both though.

IsabelLaurenceau: That makes sense. Updated the test!

selfdanielj (approved):
looks good!

PR #83, opened 2025-01-31
SRCH-5828: Indexing Crawled Data Directly into Elasticsearch

igoristic:
This pull request introduces the capability to index crawled URLs and HTML content directly into Elasticsearch, bypassing the `/urls` endpoint. This enhancement allows us to leverage our scraped data and reduce reliance on the expensive Bing API.

**Key Changes:**

- Added functionality to index scraped data into Elasticsearch
- Extracted legible "article" type information from the raw html content
- NLP: Only summary and keywords (for now)
- Removed reliance on the `/urls` endpoint (it's still an option though)

**New Dependencies:**
The following packages have been added to support Elasticsearch indexing and text processing:
```
elasticsearch==8.17.1
lxml_html_clean==0.4.1  # Required by newspaper4k
typing_extensions==4.12.2 # Required by newspaper4k
newspaper4k==0.9.3.1    # Includes nltk
cchardet==2.2.0a2      # C version of chardet
tldextract==5.1.3
pytest-asyncio==0.25.3
```

---

**Local Testing Instructions:**

**Step 1 - Set Environment Variables:**  Before running the crawler locally, set the required environment variables.  These are defined in the set_evars.sh script.  Execute the following commands to set them:
```bash
chmod +x ./set_evars.sh && source ./set_evars.sh
```

**Step 2 - Start Elasticsearch:**  Start a local Elasticsearch cluster using Docker Compose:
```bash
docker-compose up -d
```

**Step 3 - Run the Crawler:**
```bash
python -m venv venv
. venv/bin/activate
pip install -r ./search_gov_crawler/requirements.txt
python ./search_gov_crawler/benchmark.py -u \"https://www.travel.dod.mil/\" -d \"travel.dod.mil\"
```

**Step 4 - Check Elasticsearch:** (after a while since the batch size is 50)
```bash
curl --location 'http://localhost:9200/i14y-documents*/_search' \
--header 'Content-Type: application/json' \
--data '{
    "query": {
        "function_score": {
            "random_score": {
                "seed": 3,
                "field": "_seq_no"
            }
        }
    }
}'
```

#### Functionality Checks

---


search_gov_crawler/elasticsearch/convert_html_i14y.py
```diff
@@ -0,0 +1,73 @@
...
+        return None
+
+    time_now_str = current_utc_iso()
+    path = article.url or url
+
+    basename, extention = get_base_extension(url)
+    sha_id = generate_url_sha256(path)
+
+    i14y_doc = {
+        "audience": None,
```
stevenbarragan, line 23: I wonder where we get his from currently.

search_gov_crawler/elasticsearch/convert_html_i14y.py
```diff
@@ -0,0 +1,73 @@
...
+        "mime_type" : "text/html",
+        "path" : path,
+        "promote" : None,
+        "searchgov_custom1" : None,
+        "searchgov_custom2" : None,
+        "searchgov_custom3" : None,
+        "tags" : article.tags or article.keywords or article.meta_keywords,
+        "updated_at" : time_now_str,
+        "updated" : None,
+        "title_en" : article.title,
```
stevenbarragan, line 42: This should match the language. e.g. in Chinese
```
 "title_chs" : article.title
```

search_gov_crawler/elasticsearch/convert_html_i14y.py
```diff
@@ -0,0 +1,73 @@
...
+        "language" : article.meta_lang or "en",
+        "mime_type" : "text/html",
+        "path" : path,
+        "promote" : None,
+        "searchgov_custom1" : None,
+        "searchgov_custom2" : None,
+        "searchgov_custom3" : None,
+        "tags" : article.tags or article.keywords or article.meta_keywords,
+        "updated_at" : time_now_str,
+        "updated" : None,
```
stevenbarragan, line 41: Is it possible to get this from the article?

search_gov_crawler/elasticsearch/convert_html_i14y.py
```diff
@@ -0,0 +1,73 @@
...
+        "title_en" : article.title,
+        "description_en" : description_en,
+        "content_en" : content_en,
```
stevenbarragan, line 44: Same case as title. They key name should not be hardcoded.

search_gov_crawler/elasticsearch/convert_html_i14y.py
igoristic: I think it's derived from the meta tags:
> <meta name="audience" ...> tag within the <head> section of the webpage's source code; the content within this tag will usually specify the target demographic or user group the website is designed for. 

It almost never there (especially for gov websites)

search_gov_crawler/elasticsearch/convert_html_i14y.py
igoristic: Good catch! What should my backup be if the language is undetermined? I noticed we also have fields: `title` and `content` can I use those (if we don't know the language)?

search_gov_crawler/elasticsearch/convert_html_i14y.py
igoristic: Not really. The only date it has is `article.publish_date` which is not really the same thing. Though, I can write a custom extractor since we do have the raw html

search_gov_crawler/elasticsearch/install_nltk.py
```diff
@@ -0,0 +1,26 @@
+import nltk
```
selfdanielj, line 1: Do we need nltk added to the requirements file?  if you are importing here i guess we must?

search_gov_crawler/requirements.txt
```diff
@@ -11,3 +11,8 @@ scrapyd==1.5.0
 scrapyd-client==2.0.0
 scrapydweb @ git+https://github.com/GSA/searchgov-scrapydweb
 spidermon[monitoring]==1.23.0
+
+newspaper4k
+lxml_html_clean
+typing_extensions
+elasticsearch
```
selfdanielj, line 18: can you version lock these please?  also i don't see `lxml_html_clean` or `typing_extension` explicitly used, maybe they can be removed?

search_gov_crawler/search_gov_spiders/pipelines.py
```diff
@@ -12,6 +12,9 @@
 
 from search_gov_crawler.search_gov_spiders.items import SearchGovSpidersItem
 
+from search_gov_crawler.elasticsearch.es_batch_upload import SearchGovElasticsearch
+
+SPIDER_INDEX_TO_ELASTICSEARCH = os.environ.get("SPIDER_INDEX_TO_ELASTICSEARCH", True)
```
selfdanielj, line 17: Do we want this is ansible?

search_gov_crawler/elasticsearch/install_nltk.py
selfdanielj: Nevermind, looks like this package is a dependency of newspaper.

search_gov_crawler/elasticsearch/es_batch_upload.py
```diff
@@ -0,0 +1,96 @@
...
+class SearchGovElasticsearch:
+    def __init__(self, batch_size: int = 100):
+        self._current_batch = []
+        self._batch_size = batch_size
+        self._es_client = None
+        self._env_es_host = os.environ.get("ELASTICSEARCH_HOST")
+        self._env_es_port = int(os.environ.get("ELASTICSEARCH_PORT", 9200))
+        self._env_es_index = os.environ.get("ELASTICSEARCH_INDEX")
+        self._env_es_username = os.environ.get("ELASTICSEARCH_USERNAME")
+        self._env_es_password = os.environ.get("ELASTICSEARCH_PASSWORD")
```
selfdanielj, line 20: Can you change these to match the values in parameter store? It would save some complexity in the fetch and apply step if I don't have to rename them.
ELASTICSEARCH_HOST --> ES_HOSTS
ELASTICSEARCH_USERNAME --> ES_USER
ELASTICSEARCH_PASSWORD --> ES_PASSWORD

For ES_HOSTS the value in parameter store is now something like `https://host1.dev:9200,https://host2.dev:9200` and there is not a separate port variable.  I see that won't work with what you have.  How would you like these values? I can apply some sed logic to make it easier to load here.

What value do you want in ELASTICSEARCH_INDEX? Can you rename it ES_INDEX?  What types of values do you want in this one?

selfdanielj: I was able to quickly test a connect from dev spider to dev ES using these args so maybe this will work for you?  Ill plan on just passing the comma separated list and you can split it on comma to pass into `hosts` and I guess maybe you don't need the port as its own thing? Let me know if this works.

```python
es = Elasticsearch(
    hosts=["https://host1.dev:9200","https://host2.dev:9200"],
    verify_certs=False,
    http_auth=("fake-username","fake-pw")
)
```

search_gov_crawler/search_gov_spiders/pipelines.py
selfdanielj: i can open a PR for this one.

search_gov_crawler/elasticsearch/es_batch_upload.py
selfdanielj: what are the values you are expecting for ES_INDEX in the different environments?

igoristic: Will resolve this for now, since we've agreed on the index name externally

search_gov_crawler/requirements.txt
igoristic: I left some comments in `requirments.txt` that should address your concerns

search_gov_crawler/elasticsearch/convert_html_i14y.py
stevenbarragan: Can we set it up even though it's almost never there?

selfdanielj:
Is this meant to not support #85? It still has references to SPIDER_INDEX_TO_ELASTICSEARCH which will not exist any longer according to GSA/searchgov-tf#95 and GSA/searchgov-ansible#208.  

That's fine if so but i just want to know if that should be my assumption.

IsabelLaurenceau:
> Is this meant to not support #85? It still has references to SPIDER_INDEX_TO_ELASTICSEARCH which will not exist any longer according to [GSA/searchgov-tf#95](https://github.com/GSA/searchgov-tf/pull/95) and [GSA/searchgov-ansible#208](https://github.com/GSA/searchgov-ansible/pull/208).
> 
> That's fine if so but i just want to know if that should be my assumption.

Similar to Dan's. Are you planning on a separate PR that uses the field we discussed and removes SPIDER_INDEX_TO_ELASTICSEARCH?

selfdanielj:
I can't run the benchmark.py script as described from in the testing steps, after the merge of Isabel's changes.  I had to add a new argument
```python
    parser.add_argument(
        "-o",
        "--output_target",
        type=str,
        default="endpoint",
        help="Point the output of the crawls to a backend",
    )
```
and then add a benchmark arg:
```python
    if no_input_arg:
        benchmark_args = {
            "allow_query_string": args.allow_query_string,
            "allowed_domains": args.allowed_domains,
            "starting_urls": args.starting_urls,
            "handle_javascript": args.handle_js,
            "output_target": args.output_target,  # this is the new arg
            "runtime_offset_seconds": args.runtime_offset,
        }
```

search_gov_crawler/elasticsearch/es_batch_upload.py
```diff
@@ -0,0 +1,136 @@
...
+        ]
+
+    async def _batch_elasticsearch_upload(self, docs: List[Dict[Any, Any]], loop):
+        """
+        Perform bulk upload asynchronously using ThreadPoolExecutor.
+        """
+        def _bulk_upload():
+            try:
+                actions = self._create_actions(docs)
+                helpers.bulk(self._get_client(), actions)
```
selfdanielj, line 132: I think it would be helpful to log how many documents are loaded into Elastic on each upload.  We do something similar when sending data to the `/urls` endpoint.

search_gov_crawler/elasticsearch/es_batch_upload.py
```diff
@@ -0,0 +1,136 @@
...
+
+# Setup Logging
+logging.basicConfig(level=os.environ.get("SCRAPY_LOG_LEVEL", "INFO"))
+log = logging.getLogger("search_gov_crawler.es_batch_upload")
+
+if not log.handlers:
+    handler = logging.StreamHandler()
+    handler.setFormatter(JsonFormatter(fmt=LOG_FMT))
+    log.addHandler(handler)
+
```
selfdanielj, line 21: With this config I am seeing duplicate log messages from the `search_gov_crawler.es_batch_upload` logger.  I think the log config should mimic what is being done in `benchmark.py` and `scrapy_scheduler.py`:
```python
# Setup Logging
logging.basicConfig(level=os.environ.get("SCRAPY_LOG_LEVEL", "INFO"))
logging.getLogger().handlers[0].setFormatter(JsonFormatter(fmt=LOG_FMT))
log = logging.getLogger("search_gov_crawler.es_batch_upload")
```

search_gov_crawler/elasticsearch/es_batch_upload.py
selfdanielj: this would actually be better if it was logged using a spider logger so we can agregate by domain.

search_gov_crawler/search_gov_spiders/pipelines.py
```diff
@@ -28,8 +31,9 @@ def __init__(self):
         self.file_number = 1
         self.file_path = None
         self.current_file = None
+        self.es = SearchGovElasticsearch()
```
selfdanielj, line 34: could this be moved inside a conditional?  We have the file opening inside one, seems like we don't need to open a connection to ES if the spider is going to do the post to the URL endpoint, right?

search_gov_crawler/search_gov_spiders/helpers/encoding.py
```diff
@@ -0,0 +1,33 @@
...
+logging.getLogger("cchardet").addHandler(logging.NullHandler())
+
+def is_utf8_encoded(data: bytes) -> bool:
+    """Check if a byte string is UTF-8 encoded."""
+    try:
+        data.decode("utf-8")
+        return True
+    except UnicodeDecodeError:
+        return False
+
```
selfdanielj, line 15: this seems like extra, not really the python "ask for forgiveness..." model... why not just try to decode as utf-8 right away in `decode_http_response` and if you get a `UnicodeDecodeError` then carry on detecting? This way you wouldn't need the `is_utf8_encoded` function or the first conditional in `detect_encoding`

selfdanielj:
can you add the new dependencies to the `.github/dependabot.yml` so they get added to the auto-update production group?  The pytest plugin should be good but the others will needed to be called out here.
```yaml
      production-dependencies:
        patterns:
          - "python-json-logger"
          - "scrapy*"
          - "spidermon*"
```

igoristic:
@selfdanielj  @IsabelLaurenceau I added the Bing URLs to this pull-request and addressed all feedback/conflicts

search_gov_crawler/elasticsearch/es_batch_upload.py
```diff
@@ -0,0 +1,126 @@
...
+        """
+        Lazily initializes the Elasticsearch client.
+        """
+        if not self._es_client:
+            self._es_client = Elasticsearch(
+                hosts=self._parse_es_urls(self._env_es_hosts),
+                verify_certs=False,
+                http_auth=(self._env_es_username, self._env_es_password)
+            )
+            self._create_index_if_not_exists()
```
selfdanielj, line 73: `_create_index_if_not_exists` does not take a spider argument here but it needs to.  This breaks without it.  Same for `_get_client` after this change.

I wonder at this point would it make sense to have spider be a instance variable available to the whole class? Since its used by practically every function?

tests/search_gov_spiders/test_elasticsearch.py
```diff
@@ -0,0 +1,127 @@
...
+@pytest.mark.asyncio(loop_scope="module")
+async def test_batch_upload(mock_convert_html, mock_asyncio_loop):
+    es_uploader = SearchGovElasticsearch(batch_size=2)
+    mock_convert_html.return_value = {"_id": "1", "title": "Test Document"}
+    es_uploader._current_batch = [{"_id": "1", "title": "Test Document"}, {"_id": "2", "title": "Test Document"}]
+
+    es_uploader.batch_upload()
+    assert len(es_uploader._current_batch) == 0
+
+    mock_asyncio_loop.ensure_future.assert_called_once()
```
selfdanielj, line 91: this test does not run... `es_uploader.batch_uplad()` should take a spider argument so this would error if run.

To verify try running `pytest tests/search_gov_spiders/test_elasticsearch.py -k test_batch_upload`

tests/search_gov_spiders/test_elasticsearch.py
```diff
@@ -0,0 +1,127 @@
...
+@pytest.mark.asyncio(loop_scope="module")
+async def test__batch_elasticsearch_upload(mock_convert_html, mock_es_client, mock_asyncio_loop):
+    es_uploader = SearchGovElasticsearch(batch_size=2)
+    docs = [{"_id": "1", "title": "Test Document"}]
+    mock_convert_html.return_value = docs[0]
+
+    await es_uploader._batch_elasticsearch_upload(docs, mock_asyncio_loop)
+    mock_es_client.indices.create.assert_called_once()
+    mock_es_client.bulk.assert_called_once()
+
```
selfdanielj, line 110: this test also doesn't run:

`pytest tests/search_gov_spiders/test_elasticsearch.py -k test__batch_elasticsearch_upload`

.circleci/config.yml
```diff
@@ -16,6 +16,11 @@ jobs:
             python -m venv venv
             . venv/bin/activate
             pip install -r ./search_gov_crawler/requirements.txt
+      - run:
+          name: Install NLTK resources
+          command: |
+            . venv/bin/activate
+            python ./search_gov_crawler/elasticsearch/install_nltk.py
```
selfdanielj, line 23: can you add something in the readme about installing this? similar to what we have for playwright?

search_gov_crawler/elasticsearch/es_batch_upload.py
igoristic: Yeah, that's a good point. Though, is `spider` a unique instance per `item`?

search_gov_crawler/benchmark.py
```diff
@@ -39,6 +40,8 @@
 from search_gov_crawler.search_gov_spiders.extensions.json_logging import LOG_FMT
 from search_gov_crawler.search_gov_spiders.utility_files.crawl_sites import CrawlSites
 
+load_dotenv()
```
selfdanielj, line 43: Did you add these just for east of development?  Or were you thinking we should start using .env files in higher environments?

If we are officially adopt `dotenv` I think we should update the main readme with instructions on how to use since `.env` is already excluded by our main .gitigore.

search_gov_crawler/elasticsearch/es_batch_upload.py
selfdanielj: no but it is per pipeline.

selfdanielj: idk, seemed like pipeline class instance variable would work but you'd have to try it.

search_gov_crawler/benchmark.py
igoristic: @selfdanielj Yeah, this is mostly for dev/internal purposes. This makes it so we don't have to always manually set our env variables when testing. My `.env` file looks like this:
```
SPIDER_ES_INDEX_NAME="i14y-documents-spider"
SPIDER_ES_INDEX_ALIAS="i14y-documents"
ES_HOSTS="http://localhost:9200"
ES_USER="username"
ES_PASSWORD="password"
SPIDER_INDEX_TO_ELASTICSEARCH="true"
SPIDER_URLS_API="https://fake-api.com/urls"
```

igoristic: It does not effect anything production wise, since it'll just silently fail if the `.env` file does not exist

tests/search_gov_spiders/test_encoding.py
```diff
@@ -0,0 +1,22 @@
...
+    assert detect_encoding(some_str_bytes) == "ASCII" # subset of utf-8
+    assert detect_encoding(b"") == None
+
+def test_decode_http_response_utf8():
+    response_bytes = b"This is a UTF-8 string"
+    decoded_string = decode_http_response(response_bytes)
+    assert decoded_string == "This is a UTF-8 string"
+
+def test_decode_http_response_non_utf8():
+    response_bytes = b"\xc2\xa0This is not UTF-8"  # Example non-UTF-8
```
selfdanielj, line 14: this does not throw a `UnicodeDecodeError`

```python
response_bytes = b"\xc2\xa0This is not UTF-8"
response_bytes.decode("utf-8")
```

maybe you should explicitly encode as "utf-16":
```python
# change this line
response_bytes = b"\xc2\xa0This is not UTF-8"  # Example non-UTF-8`

# to something like this
non_utf8_response = "This is not UTF-8".encode("utf-16")`

# then this will raise the UnicodeDecodeError error and follow the detect_encoding path in decode_http_response
non_utf8_response.decode("utf-8")
```

search_gov_crawler/benchmark.py
selfdanielj: sure, i think it would be helpful to call this out in the readme saying we support .env files and where to place them

tests/search_gov_spiders/test_crawl_sites.py
selfdanielj, line 1: I think the new file crawl-sites-scrutiny.json should, at a minimum be added the the list of files in `test_crawl_sites_file_is_valid` but perhaps, if we are planning to have more files in the future, we could parameterize the test so that it runs on any crawl-sites*.json file in that directory... something like:

```python
def valid_crawl_sites_files() -> list[Path]:
    """Utility function to generate a list of crawl sites files in the utility_files folder"""
    utility_files = Path(__file__).parent.parent.parent.joinpath("search_gov_crawler/search_gov_spiders/utility_files/")

    return list(utility_files.glob("crawl-sites*.json"))


@pytest.mark.parametrize("crawl_sites_file", valid_crawl_sites_files())
def test_crawl_sites_file_is_valid(crawl_sites_file):
    """
    Read in the actual crawl-sites-sample.json file and instantiate as a CrawlSites class.  This will run all built-in
    validations and hopefully let you know if the file is invalid prior to attempting to run it in the scheduler.
    Additionally, we are assuming that there is at least one scheduled job in the file.
    """
    cs = CrawlSites.from_file(file=crawl_sites_file)
    assert len(list(cs.scheduled())) > 0
```

search_gov_crawler/search_gov_spiders/utility_files/csv_to_json.py
```diff
@@ -0,0 +1,68 @@
...
+    
+    with input_path.open(mode='r', encoding='utf-8') as file:
+        reader = csv.DictReader(file)
+        domains = [row["x"].strip() for row in reader if row.get("x")]
+    
+    schedules = generate_cron_schedules(start_time="01:00 FRI", count=len(domains), minute_interval=1)
+    
+    new_entries = [
+        {
+            "name": f"modules.BWEB affiliate: {domain}",
```
selfdanielj, line 45: where did you get modules.BWEB? is that from the search app?

search_gov_crawler/search_gov_spiders/utility_files/csv_to_json.py
```diff
@@ -0,0 +1,68 @@
...
+    start_day_index = weekdays.index(day_part.upper())
+    
+    for i in range(count):
+        total_minutes = i * minute_interval
+        new_dt = start_dt + timedelta(minutes=total_minutes)
+        new_day_index = (start_day_index + (new_dt.day - 1)) % 7
+        cron_schedules.append(f"{new_dt.minute:02} {new_dt.hour:02} * * {weekdays[new_day_index]}")
+    
+    return cron_schedules
+
```
selfdanielj, line 29: Does this scheduling help some problem you encountered or is this scheduling just mean to spread them out? Why not just schedule them all at once or in chunks of 10, 50, or whatever at the same time?

search_gov_crawler/search_gov_spiders/utility_files/csv_to_json.py
igoristic: This is what we used to get the bing affiliates, eg:
```
POST /human-logstash-*/_search
{
  "size": 0,
  "query": {
    "bool": {
      "filter": {
        "term": {
          "modules": "BWEB"
        }
      }
    }
  },
  ...
```

So, I figured it would be useful to know their potential source

search_gov_crawler/search_gov_spiders/utility_files/csv_to_json.py
igoristic: I ran into some thread lock issues when doing that, but maybe it's only scoped to my local machine. Never the less, I thought it would be better to gradually introduce them rather than starting multiple at once

tests/search_gov_spiders/test_crawl_sites.py
igoristic: Figured it's irrelevant, since `crawl-sites-scrutiny.json` is a subset of `crawl-sites-production.json`. But, I see your point, since we might forget to run csv to json process

tests/search_gov_spiders/test_encoding.py
igoristic: It shouldn't throw an error, since it has a silent default `str(response_bytes)`:
```
 try:
    return response_bytes.decode(detected_encoding)
except (UnicodeDecodeError, TypeError):
    return str(response_bytes)
```

IsabelLaurenceau (approved):

selfdanielj (approved):
looks good, we are merging to do testing and will apply any fixes as follow on PRs

PR #99, opened 2025-02-12
SRCH-5832: Clean item after processing

igoristic:
## Summary
Safe removing/deleting the "html_content" and "output_target" fields (and values) from the Item

### Checklist

#### Functionality Checks

#### Process Checks

### DIFF
--- a/search_gov_crawler/search_gov_spiders/pipelines.py
+++ b/search_gov_crawler/search_gov_spiders/pipelines.py
@@ -14,6 +14,10 @@ from search_gov_crawler.search_gov_spiders.items import SearchGovSpidersItem
 from search_gov_crawler.elasticsearch.es_batch_upload import SearchGovElasticsearch
 
 def safe_del(item, key: str):
+    """
+    This method prevents any exception errors if item does not have the key or is null.
+    This is just in case, since the item should always have the keys we delete
+    """
     try:
         del item[key]
     except Exception as _:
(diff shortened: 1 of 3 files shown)


IsabelLaurenceau (approved):

search_gov_crawler/search_gov_spiders/pipelines.py
```diff
@@ -13,6 +13,12 @@
 from search_gov_crawler.search_gov_spiders.items import SearchGovSpidersItem
 from search_gov_crawler.elasticsearch.es_batch_upload import SearchGovElasticsearch
 
+def safe_del(item, key: str):
+    try:
+        del item[key]
+    except Exception as _:
```
selfdanielj, line 19: Can you add a docstring to `safe_del`? 

Are you meaning to catch all Exceptions or just KeyError?  Why would the item not have the key?  Just making it safe for any delete operation?

igoristic: All exceptions, since the item might also be None. This should never happen, but in the off chance we process the same item again which will not have those keys

igoristic: @selfdanielj Added the requested doc string

selfdanielj (approved):
i did some testing to confirm items are still getting written out before and after. Looks good!  I also ran a profiler for before and after but didn't notice much difference in memory or CPU usage.... idk.  Not sure if im doing the properly or not but anyway doesn't seem to do any harm.

PR #101, opened 2025-02-19
Scrapy html parsing backup (for newspaper 4k)

igoristic:
## Summary
- Brief summary of the changes included in this PR
- Any additional information or context which may help the reviewer

### Checklist

#### Functionality Checks

#### Process Checks

### DIFF
--- a/search_gov_crawler/elasticsearch/convert_html_i14y.py
+++ b/search_gov_crawler/elasticsearch/convert_html_i14y.py
@@ -6,6 +6,7 @@ from datetime import datetime, timezone
 from urllib.parse import urlparse
 
 import search_gov_crawler.search_gov_spiders.helpers.content as content
+from search_gov_crawler.elasticsearch.parse_html_scrapy import convert_html_scrapy
 
 
 ALLOWED_LANGUAGE_CODE = {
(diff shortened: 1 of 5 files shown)


search_gov_crawler/elasticsearch/parse_html_scrapy.py
```diff
@@ -0,0 +1,82 @@
+from scrapy import Selector
+import re
+
+def clean_str(s: str) -> str:
+    if not s:
+        return s
+    return re.sub(r"\s+", " ", s).strip()
```
selfdanielj, line 7: i don't think you need this, isn't this the same as `replace_whitepsace` in `search_gov_crawler/search_gov_spiders/helpers/content.py`

search_gov_crawler/elasticsearch/convert_html_i14y.py
```diff
@@ -6,6 +6,7 @@
 from urllib.parse import urlparse
 
 import search_gov_crawler.search_gov_spiders.helpers.content as content
+from search_gov_crawler.elasticsearch.parse_html_scrapy import convert_html_scrapy
 
 
 ALLOWED_LANGUAGE_CODE = {
```
selfdanielj, line 12: i know you didn't change the ALLOWED_LANGUAGE_CODE in this PR but it seems unnecessary to have it be a dict comprehension, why not just have it be a set or list?  seems like the check below `if article.meta_lang in ALLOWED_LANGUAGE_CODE` would work just the same.

IsabelLaurenceau (approved):

search_gov_crawler/elasticsearch/parse_html_scrapy.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    meta_values = {}
+    
+    for name in meta_names:
+        value = html_selector.xpath(f"//meta[@content and (@name=\"{name}\" or @property=\"{name}\")]/@content").get()
+        meta_values[name] = value if value else None
+    
+    return meta_values
+
+def convert_html_scrapy(html_content: str) -> dict:
```
selfdanielj, line 45: More of just a comment, not suggesting a change now.

I could see this (and the i14y doc creation getting more complex and this getting harder to maintain... seems like it would be a good candidate for a class with some better organization to tie it all together.

selfdanielj (approved):
Seems to be an improvement to me, in some local tests i'm not getting any warnings that I would get before.  Nice job. Good test coverage too!

search_gov_crawler/elasticsearch/convert_html_i14y.py
igoristic: I don't think it makes much of a difference, since it only creates it once, and it's a really fast operation. I didn't want to have it static, since it would be messy with all the ...: True statements. We can use a list (or a set), but the "in" operation will no longer use a hashing function and literally loop through the list until it finds the value

search_gov_crawler/elasticsearch/parse_html_scrapy.py
igoristic: Yeah, if that's the case I will make it more modular as a single class. But, I don't think it will change much (or get any more complex) after this, since that's the store with the i14y service in general (a kind of set and forget solution)

PR #104, opened 2025-02-24
SRCH-5850: Improved elasticsearch index/alias management to accommodate the third search engine

igoristic:
Added more flexibility around elasticsearch index checking/creating/modifying, so it is more compatible with the third search engine option. Once the third search engine is ready we will only need to change two environment variables, eg:
```
SPIDER_ES_INDEX_NAME="spider-production-i14y-documents..."
SPIDER_ES_INDEX_ALIAS=""
```

### Checklist

#### Functionality Checks

#### Process Checks

### DIFF
--- a/search_gov_crawler/benchmark.py
+++ b/search_gov_crawler/benchmark.py
@@ -39,6 +39,7 @@ from pythonjsonlogger.json import JsonFormatter
 from search_gov_crawler import scrapy_scheduler
 from search_gov_crawler.search_gov_spiders.extensions.json_logging import LOG_FMT
 from search_gov_crawler.search_gov_spiders.utility_files.crawl_sites import CrawlSites
+from search_gov_crawler.elasticsearch.es_batch_upload import SearchGovElasticsearch
 
 load_dotenv()
 
@@ -121,7 +122,7 @@ def benchmark_from_file(input_file: Path, runtime_offset_seconds: int):
             **crawl_site.to_dict(exclude=("schedule",)),
         )
         scheduler.add_job(**apscheduler_job, jobstore="memory")
-
+    SearchGovElasticsearch.create_index = True
     scheduler.start()
     time.sleep(runtime_offset_seconds + 2)
     scheduler.shutdown()  # this will wait until all jobs are finished
@@ -164,7 +165,7 @@ def benchmark_from_args(
     scheduler = init_scheduler()
     apscheduler_job = create_apscheduler_job(**apscheduler_job_kwargs)
     scheduler.add_job(**apscheduler_job, jobstore="memory")
-
+    SearchGovElasticsearch.create_index = True
     scheduler.start()
     time.sleep(runtime_offset_seconds + 2)
     scheduler.shutdown()  # wait until all jobs are finished
--- a/search_gov_crawler/elasticsearch/es_batch_upload.py
+++ b/search_gov_crawler/elasticsearch/es_batch_upload.py
@@ -17,6 +17,10 @@ logging.getLogger("elastic_transport.transport").setLevel("ERROR")
 class SearchGovElasticsearch:
     """Defines the shape and methods of the spider's connection to Elasticsearch"""
 
+    # SearchGovElasticsearch.create_index is a static variable that will be set to False
+    # after each check imitating lazy singleton pattern
+    create_index: bool = False
+
     def __init__(self, batch_size: int = 50):
         self._current_batch = []
         self._batch_size = batch_size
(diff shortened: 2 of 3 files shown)


search_gov_crawler/benchmark.py
```diff
@@ -121,7 +122,7 @@ def benchmark_from_file(input_file: Path, runtime_offset_seconds: int):
             **crawl_site.to_dict(exclude=("schedule",)),
         )
         scheduler.add_job(**apscheduler_job, jobstore="memory")
-
+    SearchGovElasticsearch.create_index = True
```
selfdanielj, line 125: I tried:
* Deleted my index
* Set `SearchGovElasticsearch.create_index = False`
* Ran a benchmark

The documents were created... so i guess the index was created but i did not see the logs... looking at _settings it must have used the default because I see `"number_of_shards":"1"` not 6.

search_gov_crawler/elasticsearch/es_batch_upload.py
```diff
@@ -17,6 +17,10 @@
 class SearchGovElasticsearch:
     """Defines the shape and methods of the spider's connection to Elasticsearch"""
 
+    # SearchGovElasticsearch.create_index is a static variable that will be set to False
+    # after each check imitating lazy singleton pattern
+    create_index: bool = False
```
selfdanielj, line 22: what if this started as true so it would always check once and then you wouldn't ahve to set it in benchmark or scrapy_scheduler?

search_gov_crawler/elasticsearch/es_batch_upload.py
```diff
@@ -78,27 +82,48 @@ def _get_client(self, spider: Spider):
                 ssl_show_warn=False,
                 basic_auth=(self._env_es_username, self._env_es_password),
             )
-            self._create_index_if_not_exists(spider)
+            if SearchGovElasticsearch.create_index:
+                self._create_index_if_not_exists(spider)
+            SearchGovElasticsearch.create_index = False
```
selfdanielj, line 87: since we are not raising the error while creating the index, should this be inside the `_create_index_if_not_exists` method? I'm wondering if there is an error in creating the index we don't want to mark this as False, right?

The more i think about it, if we can't create the index we should probably just raise the error so that it stops the spider? what do you think?  I could see us not wanting to stop the spider if we get an error on one POST of documents but here I think might be a good place to choose not to contnue

search_gov_crawler/elasticsearch/es_batch_upload.py
igoristic: I don't think it would work since these will be completely separated/isolated instances of `SearchGovElasticsearch`

search_gov_crawler/benchmark.py
```diff
@@ -121,7 +122,8 @@ def benchmark_from_file(input_file: Path, runtime_offset_seconds: int):
             **crawl_site.to_dict(exclude=("schedule",)),
         )
         scheduler.add_job(**apscheduler_job, jobstore="memory")
-
+    es = SearchGovElasticsearch()
+    es.create_index_if_not_exists()
```
selfdanielj, line 126: Going back to https://github.com/GSA-TTS/searchgov-spider/pull/104#discussion_r1971619146 if create_index_if_not_exists throws an error we should stop the process here. right?

search_gov_crawler/elasticsearch/es_batch_upload.py
```diff
@@ -67,38 +73,63 @@ def _parse_es_urls(self, url_string: str) -> list[dict[str, Any]]:
...
+                    except exceptions.ElasticsearchException as e:
+                        log.error(f"Error setting alias: {str(e)}")
+                else:
+                    log.info(f"No alias set for index '{index_name}'.")
+
+        except exceptions.ElasticsearchException as e:
+            log.error(f"Elasticsearch error creating/updating index: {str(e)}")
         except Exception as e:
-            spider.logger.error(f"Error creating/checking index: {str(e)}")
+            log.error(f"General error creating/updating index: {str(e)}")
```
selfdanielj, line 132: we probably want to add a `raise` to the bottom of both of these if we want errors here to stop things

IsabelLaurenceau (approved):

search_gov_crawler/benchmark.py
igoristic: Not necessarily. Since, we still have other _output_target_: `endpoint` and `csv`

search_gov_crawler/elasticsearch/es_batch_upload.py
igoristic: @IsabelLaurenceau https://github.com/GSA-TTS/searchgov-spider/pull/104#discussion_r1973939823

search_gov_crawler/elasticsearch/es_batch_upload.py
igoristic: I think we should still keep going, in case of https://github.com/GSA-TTS/searchgov-spider/pull/104#discussion_r1973939823

search_gov_crawler/elasticsearch/es_batch_upload.py
```diff
@@ -67,38 +73,63 @@ def _parse_es_urls(self, url_string: str) -> list[dict[str, Any]]:
...
+                if alias_name:
+                    try:
+                        es_client.indices.update_aliases(body={"actions": [{"add": {"index": index_name, "alias": alias_name}}]})
+                        log.info(f"Set alias '{alias_name}' for index '{index_name}'.")
+                    except exceptions.ElasticsearchException as e:
+                        log.error(f"Error setting alias: {str(e)}")
+                else:
+                    log.info(f"No alias set for index '{index_name}'.")
+
+        except exceptions.ElasticsearchException as e:
```
selfdanielj, line 129: exceptions.ElasticsearchException is not in the 8.17 library https://elasticsearch-py.readthedocs.io/en/latest/exceptions.html... this should be something else.

search_gov_crawler/elasticsearch/es_batch_upload.py
selfdanielj, line 1: `line 150: success, _ = helpers.bulk(self._get_client(spider), actions)`

`self._get_client` no longer takes a spider argument... this needs to be removed.

igoristic: Good catch !

selfdanielj (approved):
Looks good! Nice work!

PR #100, opened 2025-02-13
SRCH-5838: Added www domain respect

igoristic:
## Summary
- Brief summary of the changes included in this PR
- Any additional information or context which may help the reviewer

### Checklist
Please ensure you have addressed all concerns below before marking a PR "ready for review" or before requesting a re-review. If you cannot complete an item below, replace the checkbox with the ⚠️ `:warning:` emoji and explain why the step was not completed.

#### Functionality Checks

- [ ] You have merged the latest changes from the target branch (usually `main`) into your branch.

- [ ] Your primary commit message is of the format **SRCH-#### \<description\>** matching the associated Jira ticket.

- [ ] PR title is either of the format **SRCH-#### \<description\>** matching the associated Jira ticket (i.e. "SRCH-123 implement feature X"), or **Release - SRCH-####, SRCH-####, SRCH-####** matching the Jira ticket numbers in the release.

- [ ] Automated checks pass. If Code Climate checks do not pass, explain reason for failures:

#### Process Checks

- [ ] You have specified at least one "Reviewer".

PR #105, opened 2025-02-24
SRCH-5838: Support domain www prefix if exists

igoristic:
## Summary
This pull-request is a replacement for: https://github.com/GSA-TTS/searchgov-spider/pull/100, which should respects the URL's www prefix if it was initially parsed with it

### Checklist

#### Functionality Checks

#### Process Checks


search_gov_crawler/elasticsearch/convert_html_i14y.py
```diff
@@ -92,8 +92,8 @@ def generate_url_sha256(url: str) -> str:
     return hashlib.sha256(url.encode()).hexdigest()
 
 def get_domain_name(url: str) -> str:
-    """Extracts the domain from a URL, removing www and ensuring consistency."""
+    """Extracts the domain from a URL, support www (only if the url was parsed with it) ensuring consistency."""
     parsed = urlparse(url if url.startswith(('http://', 'https://')) else f'https://{url}')
```
selfdanielj, line 96: this is the line that i thought it made sense to run in other places.  Shouldn't you be fixing the url prior to `urlparse` in `get_url_path` and `get_base_extension`?  Otherwise you might get different or no results for the other two fields while this one would work.

search_gov_crawler/elasticsearch/convert_html_i14y.py
```diff
@@ -92,8 +92,8 @@ def generate_url_sha256(url: str) -> str:
...
     extracted = tldextract.extract(parsed.netloc)
-    domain = f"{extracted.subdomain}.{extracted.domain}.{extracted.suffix}".lstrip('.').replace("www.", "")
+    domain = f"{extracted.subdomain}.{extracted.domain}.{extracted.suffix}".lstrip('.')
```
selfdanielj, line 98: Curious, what is the case for tldextract here vs just returning parsed.netloc?

selfdanielj: looking at the tldextract github page... why not just use `extracted.fqdn`?  Also, why do you call `urlparse` at all?  Can't you just pass in the url string to `tldextract.extract` ?

search_gov_crawler/elasticsearch/convert_html_i14y.py
igoristic: Yeah, you're absolutely right, we should ensure the http prefix everywhere to make it more consist

search_gov_crawler/elasticsearch/convert_html_i14y.py
igoristic: Yeah, looks like we don't need tldextract at all, since `urlparse(url).netloc` does the same thing

selfdanielj (approved):
domain_name looks good... this could be another ticket (or not, up to you) but im not sure that basename is getting populated when it should.  I think maybe it only gets populated if there is an extension?  Is that right?

For instance if you run fac.gov i don't see any with a basename but if you run patientcare.va.gov which has some sites that end in `.asp` you will see basenames.  Looking at the prod data I think basename should be populated more often.

IsabelLaurenceau (approved):

PR #106, opened 2025-02-25
SRCH-5852 On Disk Store

selfdanielj:
## Summary
- This implements the scrapy on-disk scheduler queue https://docs.scrapy.org/en/latest/topics/jobs.html#topics-jobs which allows us to store the queue of pending requests on disk rather than in memory and thus reducing our memory consumption and increasing stability.  We are not going to use this as a "persistent" store as discussed in the docs, it should be cleaned up on spider close and will not be used to restart a spider where it stopped at this time.
- I updated the settings.py file to enable this functionality and wrote a quick extension that cleans up the queue directory when the spider is finished.  For the folder name I chose the pid (since we are using it elsewhere) and the start time to uniquely identify the folder.
- Also, I updated the json logging to include a few more recent arguments.  When a spider logs something, that log message will now also include the `output_target` as well as a few other spider related fields.
- Finally I added new tests and adjusted existing tests so that they work with the new configuration.

### Testing
* Run a scrape job using any method (scrapy crawl / benchmark / scrapy_scheduler) and any output target.
* During the run navigate to `search_gov_crawler/jobs` and see a run specific folder containing files related to the queue.
  * If you start multiple scrapy jobs using benchmark or scrapy scheduler, each job gets its own folder.
* Once the spider(s) finishes, the run specific folder is removed.
  * This removal should happen if the spider stops for any reason including finishing the scrape, stopping due to hitting a timeout constraint, or receiving a `ctrl-C` during the run.
  * It will not get cleaned up if the process receives a kill -9 signal but since the process is immediately stopped in that case we would need another processing to come along and clean it up.  I am thinking we can address this is it becomes a problem with a cron job or something.  

### Checklist

#### Functionality Checks

#### Process Checks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -200,5 +200,6 @@ cython_debug/
 .scrapy/
 scrapy_urls/
 dbs/
+jobs/
 
 .DS_Store
--- a/search_gov_crawler/search_gov_spiders/extensions/json_logging.py
+++ b/search_gov_crawler/search_gov_spiders/extensions/json_logging.py
@@ -17,7 +17,9 @@ def search_gov_default(obj) -> dict | None:
             "name": obj.name,
             "allow_query_string": getattr(obj, "allow_query_string", None),
             "allowed_domains": getattr(obj, "allowed_domains", None),
+            "allowed_domain_paths": getattr(obj, "allowed_domain_paths", None),
             "start_urls": obj.start_urls,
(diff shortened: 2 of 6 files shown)


igoristic (approved):
Looks good! 🏆 

I was able to confirm (locally) that my memory was at a steady/consistent number (while using 5 workers). But, without this PR it was slowly climbing up

PR #109, opened 2025-02-26
SRCH-5855 Prod Errors

selfdanielj:
## Summary
- Addresses two recurring errors in prod:
  - Removes `diversity.nih.gov` from the scrutiny lists
  - Adds support for VISCII encoding
    - For the VISCII support, it appears that python does not support this but does support both unicode and cp1258 which according to wikipedia https://en.wikipedia.org/wiki/VISCII are close but may be missing some characters.  I guess if we actually try to index this we may miss some characters.  This is an "endpoint" domain however so we are just passing the URL to search anyway.  I originally had it defaulting to unicode in this case but decided on a Vietnamese-specific code page.  If we run into more issues with indexing this content I think we should revisit. 

### Testing
- For the domain removal just verify that the domain has been remove from the apropriate lists
- For the VISCII error: the page in question is https://www.accessdata.fda.gov/CMS_IA/importalert_401.html.
  - If you run `scrapy crawl domain_spider -a allowed_domains=accessdata.fda.gov -a start_urls=https://www.accessdata.fda.gov/CMS_IA/importalert_401.html -a output_target=endpoint` you should not see an error like the one in prod.
  - If you were to comment out the lines in `detect_encoding` that I have added you will see the prod error.

### Checklist

#### Functionality Checks

#### Process Checks

### DIFF
--- a/search_gov_crawler/search_gov_spiders/helpers/encoding.py
+++ b/search_gov_crawler/search_gov_spiders/helpers/encoding.py
@@ -1,28 +1,34 @@
+import logging
+import os
 
 import cchardet  # Faster and more reliable character encoding detection than chardet
-import logging
 
 # Needs muting to avoid its debug mode
-logging.getLogger("cchardet").addHandler(logging.NullHandler())
+logging.getLogger("cchardet").setLevel("INFO")
+
 
 def detect_encoding(data: bytes) -> str | None:
     """Detect the encoding of the given byte string."""
     detection_result = cchardet.detect(data)
     encoding = detection_result.get("encoding")
-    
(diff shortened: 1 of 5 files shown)


selfdanielj:
> Really just a comment I've been wondering. Why do we filter the ids in import_plist rather than just remove them from the plist. Was the plist something we compiled or we were given. (If we were given it I realize why we do the filtering and not removing it from the plist)

Yeah we were given it... i wonder, though, really, if its time to just have the JSON file be the source and delete all this conversion stuff.  If its not that time now, its probably that time soon.

PR #112, opened 2025-03-03
SRCH-5867 Enforce Output Target and Remove JS Errback

selfdanielj:
## Summary
- I noticed that i was sometimes running `scrapy crawl -a allowed_domains=example.com start_urls=https://www.example.com/` and the spider would give be weird errors, although not stop... this is because I did not include the `output_target` argument.
  - This PR makes that argument required and adds validation that it should be one of the three valid values.
  - It also moves the validation to a helper function since its the same for both spiders.
  - I added the enforcement to benchmark as well since the domain and url are already required.
- Also, I noticed when running the JS spider it could not seralize requests using the on-disk store: `Unable to serialize request: <GET http://quotes.toscrape.com/> - reason: cannot pickle 'module' object - no more unserializable requests will be logged (stats being collected)",`.  This is because we have configured a request.meta.errback callable and that is not serializable by pickle: `{'url': 'http://quotes.toscrape.com/login', 'callback': '_callback', 'errback': '_errback', 'headers': {b'Referer': [b'http://quotes.toscrape.com/js/']}, 'method': 'GET', 'body': b'', 'cookies': {}, 'meta': {'rule': 0, 'link_text': 'Login', 'playwright': True, 'errback': <bound method CrawlSpider._errback of <DomainSpiderJs 'domain_spider_js' at 0x10cf60c50>>, 'depth': 1}, 'encoding': 'utf-8', 'priority': -1, 'dont_filter': False, 'flags': [], 'cb_kwargs': {}}`.  Ultimately I don't think we were using this 

### Testing
- Ensure you cannot run a spider job without passing a `output_target` argument that is one of the three valid values.
- Ensure you can run a js spider without seeing error messages about serialization.
  - Run one of the JS domains (even quotes.toscrape.com/js) using main to see what the errors look like prior to this change.

### Checklist

#### Functionality Checks

#### Process Checks

### DIFF
--- a/search_gov_crawler/benchmark.py
+++ b/search_gov_crawler/benchmark.py
@@ -37,9 +37,10 @@ from dotenv import load_dotenv
 from pythonjsonlogger.json import JsonFormatter
 
 from search_gov_crawler import scrapy_scheduler
+from search_gov_crawler.elasticsearch.es_batch_upload import SearchGovElasticsearch
 from search_gov_crawler.search_gov_spiders.extensions.json_logging import LOG_FMT
 from search_gov_crawler.search_gov_spiders.utility_files.crawl_sites import CrawlSites
-from search_gov_crawler.elasticsearch.es_batch_upload import SearchGovElasticsearch
+from search_gov_crawler.search_gov_spiders.helpers.domain_spider import ALLOWED_CONTENT_TYPE_OUTPUT_MAP
 
 load_dotenv()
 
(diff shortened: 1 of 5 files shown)


IsabelLaurenceau (commented):
Screenshot 2025-03-04 at 8 17 26 AM

Getting correct validation for output target requirements

IsabelLaurenceau:
But I don't see the error on main "not seralize requests using the on-disk store." I'm not getting any error on your branch either though

selfdanielj:
> But I don't see the error on main that "it could not seralize requests using the on-disk store." I'm not getting any error on your branch either though

This is what I am seeing on main:
image

IsabelLaurenceau (approved):

igoristic (approved):
Looks good 🏆

PR #115, opened 2025-03-04
SRCH-5868: Removed index creation

igoristic:
The index creation/management will now be done through search-gov

We can still test elasticsearch `output_target` locally (without search-gov running) by doing:
```bash
docker-compose up
```
#### Functionality Checks

#### Process Checks

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -81,6 +81,15 @@ services:
     networks:
       - elastic
 
+  run-script:
+    image: alpine
+    env_file:
+      - .env
+    volumes:
+      - ./es_index_settings.json:/es_index_settings.json
+    command: >
+      bash -c 'if [ -f .env ]; then source .env; fi; curl -X PUT "http://es-node1:9200/$SPIDER_ES_INDEX_NAME" -H "Content-Type: application/json" -d "@es_index_settings.json"'
+
 volumes:
   es-data1:
     driver: local
--- a/es_index_settings.json
+++ b/es_index_settings.json
@@ -1,0 +1,970 @@
+{
+    "settings": {
+        "index": {
+            "number_of_shards": "6",
+            "analysis": {
+                "filter": {
+                    "en_protected_filter": {
+                        "keywords": [
+                            "gas",
+                            "fevs"
+                        ],
+                        "type": "keyword_marker"
+                    },
+                    "fr_stem_filter": {
+                        "name": "light_french",
+                        "type": "stemmer"
+                    },
+                    "it_stem_filter": {
+                        "name": "light_italian",
+                        "type": "stemmer"
+                    },
+                    "bn_stem_filter": {
+                        "name": "bengali",
+                        "type": "stemmer"
+                    },
+                    "de_stem_filter": {
+                        "name": "light_german",
+                        "type": "stemmer"
+                    },
+                    "en_stem_filter": {
+                        "name": "english",
+                        "type": "stemmer"
+                    },
+                    "es_protected_filter": {
+                        "keywords": [
+                            "ronaldo"
+                        ],
+                        "type": "keyword_marker"
+                    },
+                    "bigrams_filter": {
+                        "type": "shingle"
+                    },
+                    "hi_stem_filter": {
+                        "name": "hindi",
+                        "type": "stemmer"
+                    },
+                    "fi_stem_filter": {
+                        "name": "finnish",
+                        "type": "stemmer"
+                    },
+                    "ja_pos_filter": {
+                        "type": "kuromoji_part_of_speech",
+                        "stoptags": [
+                            "\u52a9\u8a5e-\u683c\u52a9\u8a5e-\u4e00\u822c",
+                            "\u52a9\u8a5e-\u7d42\u52a9\u8a5e"
+                        ]
+                    },
+                    "es_stem_filter": {
+                        "name": "light_spanish",
+                        "type": "stemmer"
+                    },
+                    "hu_stem_filter": {
+                        "name": "hungarian",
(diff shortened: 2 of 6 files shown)


es_index_settings.json
selfdanielj, line 1: where is the ultimate souce of this file? I don't see a json file in the searchgov PR from Steven? GSA/search-gov#1776

IsabelLaurenceau:
After deleting the existing index and running `docker compuse up` from search-services and `rails s` on branch https://github.com/GSA/search-gov/pull/1776 of searchgov a new index `i14y-documents-spider` was created when running `scrapy crawl domain_spider -a allowed_domains=quotes.toscrape.com -a start_urls=https://quotes.toscrape.com -a output_target=elasticsearch` but I thought the new index would be created with the name `development-i14y-documents-searchgov` from Steven's branch?

docker-compose.yml
```diff
@@ -81,6 +81,15 @@ services:
...
 
+  run-script:
+    image: alpine
+    env_file:
+      - .env
+    volumes:
+      - ./es_index_settings.json:/es_index_settings.json
+    command: >
+      bash -c 'if [ -f .env ]; then source .env; fi; curl -X PUT "http://es-node1:9200/$SPIDER_ES_INDEX_NAME" -H "Content-Type: application/json" -d "@es_index_settings.json"'
+
```
selfdanielj, line 92: I don't have a .env file... when i run docker compose up I get
env file searchgov-spider/.env not found: stat searchgov-spider/.env: no such file or directory... since .env is gitigored I don't think this would work on clone without special instructions to add a .env file

docker-compose.yml
```diff
@@ -81,6 +81,15 @@ services:
     networks:
       - elastic
 
+  run-script:
```
selfdanielj, line 84: did you look into the post_start element? that also might work instead of a totally separate service.
https://docs.docker.com/reference/compose-file/services/#post_start

docker-compose.yml
IsabelLaurenceau: Same. I manually exported SEARCHELASTIC_INDEX=development-i14y-documents-searchgov from steven's branch instructions but I don't know/think that's all we needed?

docker-compose.yml
igoristic: It's not much different than the `command` option, since I still have to wait for elasticsearch to become green/yellow. I still implemented it though, it might help avoid some funky race conditions

selfdanielj (commented):
I verified the index creation with this method but i think maybe there is some room to clean up the script... what about a dockerfile setup like they have in searchgov-services to install the plugins (if we really need them)? https://github.com/GSA/search-services/blob/main/docker-compose.yml

If the dockerfile setup works, maybe the only thing you need in the the script is the index creation?

docker-compose.yml
```diff
@@ -1,92 +1,50 @@
 services:
-  es-node1:
+  elasticsearch:
     image: docker.elastic.co/elasticsearch/elasticsearch:7.17.20
-    container_name: es-node1
+    container_name: elasticsearch
+    restart: always
```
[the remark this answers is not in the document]
igoristic: That's a good point. I thought I need it, since I'm exiting `elasticsearch` in the script which kills the container. But, looks like it only kills the terminal (and the watchers)

igoristic:
> I verified the index creation with this method but i think maybe there is some room to clean up the script... what about a dockerfile setup like they have in searchgov-services to install the plugins (if we really need them)? https://github.com/GSA/search-services/blob/main/docker-compose.yml
> 
> If the dockerfile setup works, maybe the only thing you need in the the script is the index creation?

@selfdanielj I went down that rabbit hole initially, but I then couldn't run the image I built because docker was asking me to login. I also didn't like the idea of always building the docker image when doing docker-compose, otherwise we would need to do it by hand when pulling the repo

selfdanielj (approved):
From a clean docker install (following the prune instructions, etc...) this works!  The docker piece is of course not really related to the code bit, but I was able to test with the new env vars and everything works!  So I think it could be merged.

But, if I stop the containers and then `docker compose up` again, I run into the same issue where it fails installing the the plugins (because they are already installed?), kills the elasticsearch container and dumps me out of the log tailing.  The elasticsearch container restarts but not kibana.  I have to `docker compse up` again to start the kibana container. 

Shouldn't it be that I just `docker compose up` and if its the first or 20th time I've run it, it just starts running?  I'd vote for a follow up story to fix the docker setup.

selfdanielj:
> > I verified the index creation with this method but i think maybe there is some room to clean up the script... what about a dockerfile setup like they have in searchgov-services to install the plugins (if we really need them)? https://github.com/GSA/search-services/blob/main/docker-compose.yml
> > If the dockerfile setup works, maybe the only thing you need in the the script is the index creation?
> 
> @selfdanielj I went down that rabbit hole initially, but I then couldn't run the image I built because docker was asking me to login. I also didn't like the idea of always building the docker image when doing docker-compose, otherwise we would need to do it by hand when pulling the repo

The login piece seems weird... but if we are just replicating what the searchgov services is doing (and that works) why can't we just... replicate it?

IsabelLaurenceau (dismissed):
Screenshot 2025-03-10 at 10 05 26 AM

selfdanielj (approved):
I think we should merge this

Docker worked as I expect when I did a full tear down and creation.  It also worked if I ctrl-c to stop the containers and docker compose up again to start them.  I then did the same thing again and got this error
```
Gracefully stopping... (press Ctrl+C again to force)
dependency failed to start: container elasticsearch is unhealthy
```

PR #117, opened 2025-03-05
poc work for conversion to jsonnet

selfdanielj:
### Proof of Concept for switch to Jsonnet for domain configurations.

Right now we have a plist and a csv file that convert to json using two different scripts.  I am not sure that either source will be changing so perhaps we should just make our own source.  Jsonnet should be familiar to this project given its use in Jemison but besides that this gives us another layer of control over the format and hopefully prevents errors in data entry.

* Check out the readme to see if it makes sense
* Check out the files and see if their construction makes sense and if it is easier to understand than the plist, csv, and existing json files.
* I added all the endpoint/scrutiny domains, FAC to the elasticsearch file (as a demonstration of making files with different sets of domains), and left the csv file blank (also as a demonstration that a blank file can work)
* Let me know your thoughts on any of my design decisions so far such as
  * Moving the files up in the project, i feel like they were really buried for such an important part of the project
  * The configuration of libsonnet and jsonnet files, I like the separation by output target.
  * I originally had all files in one directory but move the domain config files to a subdirectory so we have one json and one jsonnet file in the top level.
  * For simplicity I was thinking we keep the json files as part of the repo and not build them on deploy like jemison... we would just have to run the jsonnet command when we change things and check in the updated jsonnet/libsonnet and json files.
* Jsonnet automatically alphabetizes the fields by key in the output so the diff of the jsonnet generated files will be large due to that.
* install jsonnet (i used brew) if you want to try out the commands in the readme
* Read the next steps and see if that sounds good or if you think there is something to change or if this is a terrible idea.

If all looks good I create a story to:
* Delete the json, plist, and csv files in the utility_files folder and repoint the code to these files in the domains folder (minus fac)
* Add a step in the circleci build that lints and ensures the json files can be built without error
* See if we can add a jsonnetfmt check to pre-commnt
* whatever else you all think should be done.


search_gov_crawler/domains/config/domains_csv.libsonnet
```diff
@@ -0,0 +1,5 @@
+// csv domain config for export
+local DomainConfig = import 'domain_config.libsonnet';
+local output_target = 'csv';
+
+[]
```
IsabelLaurenceau, line 5: Is this supposed to contain the list like domains_endpoint.libsonnet does?

search_gov_crawler/domains/config/domains_elasticsearch.libsonnet
```diff
@@ -0,0 +1,12 @@
...
+local output_target = 'elasticsearch';
+[
+  {
+    name: 'Federal Audit Clearinghouse',
+    config: DomainConfig(allowed_domains='fac.gov',
+                         starting_urls='https://www.fac.gov/',
+                         schedule='00 06 * * TUE',
+                         output_target=output_target),
+  },
+]
```
IsabelLaurenceau, line 12: I'm assuming this is only fac.gov since that's the only one we want to start with?

search_gov_crawler/domains/crawl-sites-sample.jsonnet
```diff
@@ -0,0 +1,16 @@
...
+  {
+    name: domain.name,
+    allow_query_string: domain.config.allow_query_string,
+    allowed_domains: domain.config.allowed_domains,
+    handle_javascript: domain.config.handle_javascript,
+    schedule: domain.config.schedule,
+    output_target: domain.config.output_target,
+    starting_urls: domain.config.starting_urls,
+  }
+  for domain in endpoint_domains
```
IsabelLaurenceau, line 15: I'm wondering about separating the sample, production and production -scrutiny list like this. It seems very repetitive, can we not use environment variables to pass in what we want to create? I also wonder because sometimes I run the sites-sample as csv when I'm doing depth testing or other testing stuff.

search_gov_crawler/domains/config/domains_elasticsearch.libsonnet
IsabelLaurenceau: I see your comment in the ReadMe about "as a demonstration of making files with different sets of domains" but this would also be how we update to the batches correct?

selfdanielj: right, my idea would be as we do more batches we would add them to this file.

search_gov_crawler/domains/crawl-sites-sample.jsonnet
selfdanielj: i created three jsonnet files (sample, production, and production-scrapy) because that's what we have now... im all for making it more concise though.  What would your idea look like in practice? Would there be one jsonnet file that outputs something different per environment (variable)?  Would we still have them checked into the repo or build them for each environment?

As for testing, the jsonnet files themselves are pretty simple so maybe you could just make one that does what you need for the test? For instance:
```jsonnet
local endpoint_domains = import 'config/domains_endpoint.libsonnet';
[
  {
    name: domain.name,
    allow_query_string: domain.config.allow_query_string,
    allowed_domains: domain.config.allowed_domains,
    handle_javascript: domain.config.handle_javascript,
    schedule: domain.config.schedule,
    output_target: 'csv',
    starting_urls: domain.config.starting_urls,
  }
  for domain in endpoint_domains
]
```

selfdanielj: I did find this just now https://jsonnet.org/learning/getting_started.html#multi... we could also just have one jsonnet file and have it output multiple files... could be one per environment or something.

IsabelLaurenceau: Yeah I think 1) using the multiple file output in a single file format that you linked would be good. 2) I think if it's all in  a single file using variables would make it easier to see what's different between the sets.

selfdanielj: made updates based  on slack conversations: moved the json generation to a single files and switched to an environment based naming convention

IsabelLaurenceau (approved):
🎉

selfdanielj:
Im going to leave this open as a reference and place for future discussion until i can replace it with a PR that implements all the changes.

selfdanielj:
Moved to #120

PR #120, opened 2025-03-06
SRCH-5869 Move Domain Config to Jsonnet

selfdanielj:
## Summary
This PR moves us from a plist and csv source with code to transform them into json crawl site files, to a handful of jsonnet files that are used to generate the json crawl site files.  In the `search_gov_crawler/domains` folder you will now see a single `crawl-sites.jsonnet` file and the three json files that have been generated from it.  In the `search_gov_crawler/domains/config` folder are libsonnet files that contain the domain information, separated by output target.

**I originally had the fac.gov domain uncommented in the elasticsearch domains file but decided to comment it out in case we want to move these changes to prod prior to the third search engine being ready.  We can always undo that.**

Changes:
- Added `search_gov_crawler/domains` that contains the jsonnet and json files. 
- Deleted `utility_files` folder and all associated tests. Moved the scrutiny `.plist` and bing domains `.csv` files inside this folder just in case we need them.
- Moved `crawl_sites.py` from `utility_files` up a level and re-pointed all imports to the new location.
- Updated circleci to run tests using the `jsonnetfmt` and `jsonnet` commands to ensure proper formatting and check if a jsonnet file was changed but the json files were not updated
- Added jsonnet precommit to check formatting
- Updates to main README and docs pages
- New README inside `domains` folder

### Tests
* Install jsonnet.  I used the standard (c++) implementation but the precommit and circleci cimg use the golang implementation.  Either should be fine.
* Test out running jsonnetfmt as described in the README.
* Try changing something and see that the changes are reflected in the json files when you run the jsonnet command as described in the README.
* Run scrapy_scheduler.py with the default (produtction) file or use the SPIDER_CRAWL_SITES_FILE_NAME env var to change the file to one of the others.  Ensure proper functionality.
* Run `benchmark.py -f <one of the new json files>`. Ensure proper functionality.
* **Please review the new JSON files to make sure I didn't accidentally changes something**. The diff in github was going to be large because jsonnet automatically alphabetizes the keys.  I used the `jd` utility https://github.com/josephburnett/jd to check before and after and I think it looks good.

### Checklist

#### Functionality Checks

#### Process Checks


IsabelLaurenceau (approved):
Went through testing instructions and everything working as expected. (Also I was using the golang implementation so that can be considered tested as well) 

One comment/idea for future discussion/work: 
I did one extra test to see what would happen if we have the same domain in two different libsonnets. I added navy all hands to the domains_csv.libsonnet and ran the same jsonnet instruction. (I wanted to see do they get de-duplicated or not) They aren't deduplicated but instead one is added for each endpoint. I think possible future work could include allowing output_target to have multiple endpoints, because right now as is I believe this would cause allhands.navy.mil to be crawled twice? 
 
Screenshot 2025-03-10 at 9 04 39 AM

selfdanielj:
> Went through testing instructions and everything working as expected. (Also I was using the golang implementation so that can be considered tested as well)
> 
> One comment/idea for future discussion/work: I did one extra test to see what would happen if we have the same domain in two different libsonnets. I added navy all hands to the domains_csv.libsonnet and ran the same jsonnet instruction. (I wanted to see do they get de-duplicated or not) They aren't deduplicated but instead one is added for each endpoint. I think possible future work could include allowing output_target to have multiple endpoints, because right now as is I believe this would cause allhands.navy.mil to be crawled twice?

Interesting... i thought you were trying that as a breaking case.  Can you try to run the test suite with the duplicate in the file?  I think it might break there, I thought I added a test that tries to validate the files.  Do you see this as a valid use case?  I guess adding csv + one of the others is the only way I could see us wanting to be able to do two.

IsabelLaurenceau:
> > Went through testing instructions and everything working as expected. (Also I was using the golang implementation so that can be considered tested as well)
> > One comment/idea for future discussion/work: I did one extra test to see what would happen if we have the same domain in two different libsonnets. I added navy all hands to the domains_csv.libsonnet and ran the same jsonnet instruction. (I wanted to see do they get de-duplicated or not) They aren't deduplicated but instead one is added for each endpoint. I think possible future work could include allowing output_target to have multiple endpoints, because right now as is I believe this would cause allhands.navy.mil to be crawled twice?
> 
> Interesting... i thought you were trying that as a breaking case. Can you try to run the test suite with the duplicate in the file? I think it might break there, I thought I added a test that tries to validate the files. Do you see this as a valid use case? I guess adding csv + one of the others is the only way I could see us wanting to be able to do two.

Oh yeah I didn't run the test with the duplicates but they do fail there. I see it more as a precaution than a use case? If we add things to multiple files (i.e csv or elastic or endpoint libsonnets) the way the jsonnet is written now will duplicate it. I don't think we would ever want elastic and endpoint to be duplicated. but I could see wanting elastic with csv or endpoint with csv for testing purposes. 

Screenshot 2025-03-10 at 9 24 39 AM

igoristic (approved):
Looks good 🏆

PR #132, opened 2025-03-14
SRCH-5871: Added support for parsing PDFs (v2)

igoristic:
**Local testing:**

STEP: 1
---
```
# WARNING: Some of these cmds will delete *all* persistent docker data. Use with caution!

# stop/remove all docker images:
docker stop $(docker ps -a -q)
docker rm $(docker ps -a -q)

# remove all docker mounted/binded data:
docker system prune -a
docker volume prune -a

# start docker compose 
docker-compose up

# ...wait about a minute for docker to finish all the automation

# confirm plugins are installed:
curl --location 'http://localhost:9200/_cat/plugins'

# confirm default/dev elasticsearch index is created "development-i14y-documents-searchgov":
curl --location 'http://localhost:9200/_cat/indices'

# Now you can run spiders locally with "output_target": "elasticsearch"
# Also works if search-gov docker containers are already running
```

STEP: 2
---
Ingest a site with known pdf files, eg:
```
scrapy crawl domain_spider -a allowed_domains=cms.gov -a start_urls=https://www.cms.gov -a output_target=elasticsearch
```

STEP: 3
---
Go to Kibana > Dev Tools and confirm pdf documents are in elasticsearch via:
```
GET development-i14y-documents-searchgov/_search
{
  "sort": [
    {
      "_score": {
        "order": "desc"
      },
      "created_at": {
        "order": "desc"
      }
    }
  ],
  "size": 500,
  "_source": true,
  "query": {
    "bool": {
      "must": [
        {
          "bool": {
            "should": [
              {
                "query_string": {
                  "fields": [
                    "extension"
                  ],
                  "query": "pdf"
                }
              }
            ],
            "minimum_should_match": 1
          }
        }
      ]
    }
  }
}
```

#### Functionality Checks

#### Process Checks


search_gov_crawler/elasticsearch/convert_pdf_i14y.py
```diff
@@ -7,6 +7,52 @@
...
+        # Get all hidden links
+        page_object = page.get_object()
+        if key in page_object.keys():
+            ann = page_object[key]
+            for a in ann:
+                u = a.get_object()
+                if uri in u[ank].keys():
+                    link = u[ank][uri]
+                    links[link] = link
+    return links.values()
```
selfdanielj, line 55: Did you find this effective?  It seems overly complex just to add something back in that we stripped out.. did you look into if we could just not strip them out?

search_gov_crawler/elasticsearch/convert_pdf_i14y.py
```diff
@@ -52,14 +102,21 @@ def convert_pdf(response_bytes: bytes, url: str, response_language: str = None):
...
         "url_path": get_url_path(url),
         "domain_name": get_domain_name(url),
     }
+
+    add_title_and_filename(content_key, title_key, i14y_doc)
+    add_title_and_filename(description_key, title_key, i14y_doc)
+    all_links = get_links_set(reader)
+    i14y_doc[content_key] = f"{i14y_doc[content_key]} {" ".join(all_links) if len(all_links) > 0 else ""}"
+
+    return i14y_doc
```
selfdanielj, line 119: i don't particularly like hiding the dict manipulation in a function like this, at a glance i can't tell that those fields are getting modified.

selfdanielj:
Seems to work, I can see PDFs in the index!

I don't need any changes for my comments but I guess I'll just reiterate my concern about the complexity we are adding to the creation of the documents.  Fields go through multiple updates, multiple functions in multiple files, its getting hard to trace.  I'd just ask that as we make additions we look for ways to simplify things... maybe encapsulating all transformations for a given field into a single function? instead of having several ORs around the file? If I am getting annoyed at following some of these fields through all the options I can only imagine a greater difficulty for someone new coming to this.

selfdanielj (approved):

search_gov_crawler/elasticsearch/convert_pdf_i14y.py
igoristic: Yeah, I was trying to mimic the behavior in the original i14y. Not sure if this helps the indexing/etc. Though, I think it might be worth adding the "hidden" links

search_gov_crawler/elasticsearch/convert_pdf_i14y.py
igoristic: Not sure what you mean. This is normal functional pattern. Are there any examples/patterns we should try to follow?

IsabelLaurenceau:
> Seems to work, I can see PDFs in the index!
> 
> I don't need any changes for my comments but I guess I'll just reiterate my concern about the complexity we are adding to the creation of the documents. Fields go through multiple updates, multiple functions in multiple files, its getting hard to trace. I'd just ask that as we make additions we look for ways to simplify things... maybe encapsulating all transformations for a given field into a single function? instead of having several ORs around the file? If I am getting annoyed at following some of these fields through all the options I can only imagine a greater difficulty for someone new coming to this.

Agree. @igoristic can you create a new ticket to address some of Dan's concerns? I don't think we need to do that right now but I want them documented for future work so we don't lose them.

IsabelLaurenceau (approved):

search_gov_crawler/elasticsearch/convert_pdf_i14y.py
igoristic: I see your point, but I think it becomes more of a personal preference/style at that point

PR #134, opened 2025-03-17
SRCH-5882-Deep-Search

IsabelLaurenceau:
## Summary
Added ability to specify depth_limit in command line or in json file. 
Checks to make sure depth is between 0 and 250 (as 0 is treated as no limit and 250 is probably even further than we would need to go). 
Updated tests to take into account new search_depth argument.  

## Testing: 
1) Run all pytest tests
2) Test both using command line spider and benchmark/scheduler. 

2a) Command Line Spider
- `scrapy crawl domain_spider -a allowed_domains=quotes.toscrape.com -a start_urls=https://quotes.toscrape.com -a output_target=csv -a depth_limit=1` 
-- Should work as expected, when spider starts make sure you see DEPTH_LIMIT': '1'
 - `scrapy crawl domain_spider -a allowed_domains=quotes.toscrape.com -a start_urls=https://quotes.toscrape.com -a output_target=csv -a depth_limit=0`  & - `scrapy crawl domain_spider -a allowed_domains=quotes.toscrape.com -a start_urls=https://quotes.toscrape.com -a output_target=csv -a depth_limit=300` 
 - -- Should return `ValueError: Search Depth must be between 1 and 250 inclusive. You submitted: 0 
`

2b) python benchmark.py -f domains/crawl-sites-development.json
Should return normal spider. You can go into the json and switch the search_depth to something other than 3 and again observe that it's different when ran.

### Checklist

#### Functionality Checks

#### Process Checks


search_gov_crawler/search_gov_spiders/spiders/domain_spider.py
```diff
@@ -82,13 +91,30 @@ def __init__(
...
+            start_urls.split(",")
+            if start_urls
+            else helpers.default_starting_urls(handle_javascript=False)
         )
 
+    @classmethod
+    def from_crawler(cls, crawler, *args, **kwargs):
+        spider = super().from_crawler(crawler, *args, **kwargs)
+        if "search_depth" in kwargs:
+            if int(kwargs["search_depth"]) > 250 or int(kwargs["search_depth"]) < 1:
```
igoristic, line 109: Also, https://github.com/GSA-TTS/searchgov-spider/pull/134/files#r1999538305

search_gov_crawler/benchmark.py
```diff
@@ -231,6 +237,14 @@ def benchmark_from_args(
         default=5,
         help="Number of seconds to offset job start",
     )
+    parser.add_argument(
+        "-sd",
+        "--search_depth",
+        type=int,
+        default=3,
+        help="How far down should spider crawl",
+        choices=range(1, 250),
```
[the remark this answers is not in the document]
IsabelLaurenceau: no we don't want them to use 0

search_gov_crawler/search_gov_spiders/spiders/domain_spider_js.py
```diff
@@ -116,11 +118,30 @@ def __init__(
...
+            else helpers.default_starting_urls(handle_javascript=True)
         )
-        self.start_urls = start_urls.split(",") if start_urls else helpers.default_starting_urls(handle_javascript=True)
         self.output_target = output_target
 
+    @classmethod
+    def from_crawler(cls, crawler, *args, **kwargs):
+        spider = super().from_crawler(crawler, *args, **kwargs)
+        if "search_depth" in kwargs:
+            if int(kwargs["search_depth"]) > 250 or int(kwargs["search_depth"]) < 1:
```
igoristic, line 136: Also, https://github.com/GSA-TTS/searchgov-spider/pull/134/files#r1999538305

igoristic (dismissed):
Looks good 🥇 

Just some minor comments/clarification with min and max:

-If 0 is unlimited then why is the minimum 1?
-If we have unlimited then do we really need a max of 250?

I think it should just validate to make sure the number is not negative, eg:
```
if (value < 0):
    # handle invalid value here
```

selfdanielj:
kind of superficial but we are not configuring `search_depth` but rather `crawl_depth` or perhaps `depth_limit` right?

search_gov_crawler/scrapy_scheduler.py
```diff
@@ -66,10 +68,25 @@ def run_scrapy_crawl(
...
-        msg, spider, allow_query_string, allowed_domains, start_urls, output_target
+        msg,
+        spider,
+        allow_query_string,
+        allowed_domains,
+        start_urls,
+        output_target,
+        search_depth,
+    )
+    log.info(
```
igoristic, line 82: @IsabelLaurenceau Looks like you have an extra/duplicated `log.info(...)` function

search_gov_crawler/search_gov_spiders/spiders/domain_spider.py
```diff
@@ -82,13 +91,30 @@ def __init__(
...
-            start_urls.split(",") if start_urls else helpers.default_starting_urls(handle_javascript=False)
+            start_urls.split(",")
+            if start_urls
+            else helpers.default_starting_urls(handle_javascript=False)
         )
 
+    @classmethod
+    def from_crawler(cls, crawler, *args, **kwargs):
+        spider = super().from_crawler(crawler, *args, **kwargs)
+        if "search_depth" in kwargs:
```
selfdanielj, line 108: you can (and probably should) make search_depth (or whatever we call it) a kw argument here:
```python
    @classmethod
    def from_crawler(cls, crawler, search_depth: int, *args, **kwargs):
        spider = super().from_crawler(crawler, *args, **kwargs)
        if int(search_depth) > 250 or int(search_depth) < 1:
```
this way it is required and you don't have to go searching for it in the kwargs dict.

search_gov_crawler/search_gov_spiders/spiders/domain_spider.py
```diff
@@ -53,8 +54,15 @@ class DomainSpider(CrawlSpider):
              -a output_target=elasticsearch```
     """
 
+    DEPTH_LIMIT = os.environ.get("SPIDER_DEPTH_LIMIT", "3")
```
selfdanielj, line 57: if you remove the __init__ arg below, I don' think you need this but that got me thinking... should this setting be removed from the settings.py file all together?  Do we want there to be a default?  As-is I could set SPIDER_CRAWL_DEPTH to 0, not pass -a search_depth to scrapy crawl and it would crawl to unlimited depth.

selfdanielj: I guess this goes for the environment variable as well.  Should we remove it?

IsabelLaurenceau:
> kind of superficial but we are not configuring `search_depth` but rather `crawl_depth` or perhaps `depth_limit` right?

I didn't want to use depth_limit because of the environment variable but you mentioned in another comment how maybe we shouldn't use that environment variable to start with and I think you are right. I am going to remove the environment variable and change `search_depth` to `depth_limit`

search_gov_crawler/search_gov_spiders/spiders/domain_spider.py
IsabelLaurenceau: If we have the default in settings why would we want to make it required? Do you think that makes it more stable or is using the default as is now (with my new changes) enough?

search_gov_crawler/search_gov_spiders/spiders/domain_spider.py
selfdanielj: ok sure... maybe add a comment in the from_crawler method reminding about where the default is set? I worry that we now have a few places where it could be set so its not as clear as "always look in the settings file"

IsabelLaurenceau: Fair, comment added!

igoristic (approved):

search_gov_crawler/search_gov_spiders/spiders/domain_spider.py
selfdanielj: I guess I wasn't sure what the plan was for that when I left the comment.  Leaving it in settings makes sense to me based on what we talked about.

I still think it makes it more readable to list it as an arg with a default of None here if you know you are going to access it, rather than getting it from the kwargs dict:
```python
def from_crawler(cls, crawler, search_depth: int | None = None, *args, **kwargs):
``` 
or something?

IsabelLaurenceau: I'm not sure where I would get the search_depth to put in the arguments though? We don't actually call from_crawler it's included in the scrapy middleware

selfdanielj: i think `scrapy crawl` starts the spider you specificy using the from_crawler method and the rest of the args you specify get passed in.  Its buried in subclasses in the scrapy repo https://github.com/scrapy/scrapy/blob/872924721344502cca92a72ce68d5f78a9b9e5bc/scrapy/crawler.py#L163C1-L164C66.  Here is the actual call.

IsabelLaurenceau: Okay thank you for finding this. I'll make the change. I might need look further into the documentation so please hold :D

IsabelLaurenceau: >.< Apologies took so long because I kept testing `scrapy crawl domain_spider -a .... output_target=csv -a search_depth=2` instead of `scrapy crawl domain_spider -a .... output_target=csv -a depth_limit=2`

igoristic (approved):

selfdanielj (approved):
👍

PR #136, opened 2025-03-19
SRCH-5883: Batch 4 - Added NOAA domains

igoristic:
This adds top 70 NOAA domains to elasticsearch output.

****Note** this pull-request, should only be merged after the "depth limit" pr is merged: https://github.com/GSA-TTS/searchgov-spider/pull/134

### Checklist

#### Functionality Checks

#### Process Checks


search_gov_crawler/domains/config/domains_elasticsearch.libsonnet
```diff
@@ -48,4 +48,566 @@ local output_target = 'elasticsearch';
                          schedule='15 06 * * TUE',
                          output_target=output_target),
   },
+
+  // Batch 4 - NOAA domains:
+  {
+    name: 'NOAA (noaa)',
+    config: DomainConfig(allowed_domains='noaa.gov',
+                         starting_urls='https://www.noaa.gov/',
+                         schedule='01 18 * * WED',
```
selfdanielj, line 57: did you pick 1800 because it was coming up?  Is this the best time for these to start regularly?  I had been just starting them near the top of the day, 0600 to avoid DST issues around which day it starts.

search_gov_crawler/domains/config/domains_elasticsearch.libsonnet
```diff
@@ -48,4 +48,566 @@ local output_target = 'elasticsearch';
...
+                         output_target=output_target,
+                         depth_limit=8),
+  },
+  {
+    name: 'National Weather Service (nws)',
```
selfdanielj, line 62: Did you do any optimization for instance, making sure the longest one or two get started early and letting smaller ones run one after another while those long runs go?  Or are these just all of them together?  I guess we can optimize the schedule once we see how they run.

search_gov_crawler/domains/config/domains_elasticsearch.libsonnet
igoristic: I changed them all to `19 * * FRI`, so they can start around noon (PDT) on Friday, since it's not too late and not too early and the traffic is a lot less.

DST doesn't really apply to UTC, and it's not really important what day it starts (Thursday night or Friday AM), since we don't have any strict rules around "when" the data get's indexed

search_gov_crawler/domains/config/domains_elasticsearch.libsonnet
igoristic: Yeah, they are ordered chronologically (in a descending order) based on record count

search_gov_crawler/domains/config/domains_elasticsearch.libsonnet
selfdanielj: Ok, I don't know if we are going to be in prod with all this by Friday either.  We can always run it manually if it's past the schedule.

For DST I meant making it easy to say it's scheduled for Mondays instead of oh it's Monday except for Oct-Mar when it's late Sundays.  But you are right that it doesn't really matter when it runs.

search_gov_crawler/search_gov_spiders/crawl_sites.py
```diff
@@ -68,14 +68,12 @@ def __iter__(self):
...
-        unique_sites = {
-            (crawl_site.allowed_domains, crawl_site.starting_urls)
-            for crawl_site in self.root
-        }
-        if len(unique_sites) != len(self.root):
-            msg = "The combination of allowed_domain and starting_urls must be unique in file!"
-            raise TypeError(msg)
+        domains_map = {}
+        for site in self.root:
+            site_key = f"{site.output_target}::{site.allowed_domains}"
```
selfdanielj, line 73: why did you change the key?

search_gov_crawler/domains/config/domains_elasticsearch.libsonnet
IsabelLaurenceau: This also gave me an idea for future work. If we can parse all these schedules and create a table that we (as people) can easily see what's scheduled for when. I think that will make it so 1) it will help with optimization of schedules later on and 2) it's not a thing we manually have to keep up which will inevitably become out of sync.

search_gov_crawler/search_gov_spiders/crawl_sites.py
igoristic: According to my discussion with Jim, we can have multiple domains with different output targets, for example currently we have multiple  `coastwatch.glerl.noaa.gov` with output_target: "endpoint" **and** "elasticsearch"

search_gov_crawler/domains/config/domains_elasticsearch.libsonnet
selfdanielj: good idea! A while back i was trying to convert it into mermaid but i couldn't really find the right format for a recurring schedule like this so i gave up.  Gantt was close.

IsabelLaurenceau: Made a ticket for it in Jira: https://cm-jira.usa.gov/browse/SRCH-5903

PR #142, opened 2025-03-25
SRCH-5901: Batch 7- Air force

igoristic:
## Summary
This adds **all** the batch 7 domains with exception of `www.afwerx.af.mil` which is used by two different affiliates `af_werx` and `afpw_afwerx`, and also `bandofthesouthwest.ang.af.mil` one with `www.` prefix and one without.

**Note:**
This pull-request also includes an automation utility to convert any csv file into libsonnet (jsonnet). There's a simple example (and instructions) in `convert_libsonnet/from_csv.py` that you can test by doing:
```bash
python ./convert_libsonnet/from_csv.py 
``` 

#### Functionality Checks

#### Process Checks


convert_libsonnet/Bing Transition Batches  - Batch 7.libsonnet
selfdanielj, line 1: did you mean to leave these files here?

convert_libsonnet/from_csv.py
```diff
@@ -0,0 +1,76 @@
...
+            f.write(libsonnet_str)
+
+
+if __name__ == "__main__":
+    options = {
+        "file_name": "Bing Transition Batches  - Batch 7.csv",
+        "column_index": {
+            "name": 1,
+            "affiliate": 2,
+            "allowed_domains": 3,
```
selfdanielj, line 70: I though we were removing the "www" from the start url for the allowed domains?  You did that for your last batch.

convert_libsonnet/from_csv.py
```diff
@@ -0,0 +1,76 @@
+import csv
+import os
+from pathlib import Path
+
+def convert_to_libsonnet(options):
+    """
+    Convert a CSV file to a .libsonnet file, that can then be copy-pasted into
```
selfdanielj, line 7: can you add the columns you expect and what they contain to the docstring?

convert_libsonnet/from_csv.py
```diff
@@ -0,0 +1,76 @@
...
+
+if __name__ == "__main__":
+    options = {
+        "file_name": "Bing Transition Batches  - Batch 7.csv",
+        "column_index": {
+            "name": 1,
+            "affiliate": 2,
+            "allowed_domains": 3,
+        },
+        "depth_limit": 8,
```
selfdanielj, line 72: shouldn't depth limit respect a deep or shallow indicator?  They may all be deep here but they all aren't

convert_libsonnet/from_csv.py
```diff
@@ -0,0 +1,76 @@
...
+        "schedule": "30 20 * * FRI",
```
selfdanielj, line 73: why is this the default?  I really think we should be spreading them out more.  Sure they all just queue up and run whenever but the bad thing about having 1000 starting on Friday evening is we can't naturally restart from the middle of the list wheras if they are spread over the week if we release or have to restart on any day we can start right from whatever is supposed to run the next day instead of waiting for Friday to start the list from the top again.

selfdanielj (commented):
Nice idea with the script... left some comments.

convert_libsonnet/Bing Transition Batches  - Batch 7.libsonnet
igoristic: I can remove them, just thought it would be helpful as example. The csv column options rely on column index and not names. Let me know if I should remove these files, I'm fine either way

convert_libsonnet/from_csv.py
igoristic: You're right! removed

convert_libsonnet/from_csv.py
igoristic: I'm going off this comment (in the ticket) from Jim:
> All of these need to be a deep crawl- so like 8 hops

IsabelLaurenceau (approved):

selfdanielj (dismissed):
Not sure the big example files makes sense nor does leaving the depth limit static but those aren't blockers to me to getting the schedule in place.

igoristic:
@selfdanielj @IsabelLaurenceau  Looks like I will need another approval, since I resolved merge conflicts and addressed some of code review feedback

selfdanielj (approved):

PR #143, opened 2025-03-25
SRCH-5891 only dates

IsabelLaurenceau:
## Summary
- This only implements safe date checking for documents going to elastic search. supremecourt.gov was one of the domains throwing errors observed in prod. 

- To test this first run the updated tests. 

- Once those pass I add a print statment into conver_html_i14y.py to verify what's happening before the return statement. 
- `       print(
        f"D{article_backup['created_at']}:{parse_date_safley(article_backup['created_at'])}"
    )` 

- Spin up local elastic & kibana and run a crawl: 

 `scrapy crawl domain_spider -a allowed_domains=supremecourt.gov -a start_urls=https://www.supremecourt.gov -a output_target=elasticsearch -a depth_limit=3`     

- You should see in the output the conversion of date types. 
- Make sure documents are being added to index without date errors. 
- in dev tools in kibana (http://0.0.0.0:5601/app/dev_tools#/console) verify that for changed you are getting "changed" :"", for a running GET development-i14y-documents-searchgov/_search { "query": { "match_all": {} } }
 
### Checklist

#### Functionality Checks

#### Process Checks


search_gov_crawler/elasticsearch/convert_html_i14y.py
```diff
@@ -2,17 +2,27 @@
...
+)
 
 
 def convert_html(response_bytes: bytes, url: str, response_language: str = None):
     """Extracts and processes article content from HTML using newspaper4k."""
     html_content = encoding.decode_http_response(response_bytes=response_bytes)
     config = newspaper.Config()
     config.fetch_images = False  # we are not using images, do not fetch!
-    config.clean_article_html = False  # we are not using article_html, so don't clean it!
+    config.clean_article_html = (
```
selfdanielj, line 23: what is the line-limit you have in place for these formats?  There is searchgov-spider/pyproject.toml with pylint settings.  Are you using something else locally?  can you set it to 120?

IsabelLaurenceau: oh let me check. I did see that pylint updated last time I ran pip install

search_gov_crawler/elasticsearch/i14y_helper.py
```diff
@@ -1,10 +1,11 @@
 import re
 import os
 import hashlib
+from datetime import UTC, datetime
+from dateutil import parser
```
selfdanielj, line 5: if we are introducing a direct use of dateutil can you add this to our requirements and dependabot files?  I'm assuming its also required by scrapy or something

search_gov_crawler/elasticsearch/i14y_helper.py
```diff
@@ -29,9 +31,16 @@ def null_date(article_date):
...
 
     Returns:
-        str: Either the original date, or NoneValue
+        str: Either the date in isoformat (YYYY-MM-DD'T'HH:MM:SS), or None
     """
-    return article_date or None
+    if type(date_value) is not str:
+        return None
+    try:
+        datetime_object = parser.parse(date_value, ignoretz=True, fuzzy=True)
```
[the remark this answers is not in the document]
IsabelLaurenceau: You are correct, let me verify it still works

search_gov_crawler/elasticsearch/i14y_helper.py
```diff
@@ -29,9 +31,16 @@ def null_date(article_date):
         article_date (str | Any): Date value that needs to fallback to null
```
selfdanielj, line 31: this doc string needs the name of the arg updated from article_date to date_value

search_gov_crawler/elasticsearch/i14y_helper.py
```diff
@@ -20,18 +22,29 @@
...
     Returns:
-        str: Either the original date, or NoneValue
+        str: Either the date in isoformat (YYYY-MM-DD'T'HH:MM:SS), or None
     """
-    return article_date or None
+    if not isinstance(date_value, str):
+        return None
+    try:
+        datetime_object = parser.parse(date_value, ignoretz=False, fuzzy=True)
+        utc = pytz.timezone("UTC")
```
selfdanielj, line 41: i dont think you need pytz here, just use datetime.UTC it is already imported in this file.

IsabelLaurenceau: Yeah I actually removed all that. I'll send the re-request once I finish the changes

IsabelLaurenceau: Alright, ready for re-review :)

search_gov_crawler/elasticsearch/i14y_helper.py
```diff
@@ -20,18 +21,27 @@
 }
 # fmt: on
 
-def null_date(article_date):
+
+def parse_date_safley(date_value: any) -> str:
```
selfdanielj, line 25: is this function name a typo?  Should it be `parse_date_safely`?

IsabelLaurenceau: yes

search_gov_crawler/elasticsearch/i14y_helper.py
```diff
@@ -20,18 +21,27 @@
...
-    return article_date or None
+    if not isinstance(date_value, str):
+        return None
+    try:
+        datetime_object = parser.parse(date_value, fuzzy=True)
+        datetime_format = "%Y-%m-%dT%H:%M:%S"
+        return datetime_object.strftime(datetime_format)
+    except ValueError:
+        return None
+
```
selfdanielj, line 44: what did you think about logging something here in the case when we can't parse... i guess that means both `None` cases.

IsabelLaurenceau: So I thought about it. I don't think we should just log if we don't have any thing because some sites just might not include date data and I don't think that helps our system know anything. We could log a warning if we found a date but couldn't parse it though

selfdanielj: Yeah good call... i think you could be a bit more clear though... also what is throwing a ValueError?  The length check doesn't consider if we get an integer or something.... not sure if thats possible.  What about something like this?

```python
from typing import Any
from dateutil.parser import ParserError

def parse_date_safely(date_value: Any) -> str | None:
    """
    Convert falsey date values (like an empty string) to None,
    which will yield a null value in elasticsearch.

    Args:
        date_value (str | Any): Date value that needs to fallback to null

    Returns:
        str: Either the date in isoformat (YYYY-MM-DD'T'HH:MM:SS), or None
    """
    if date_value is None:
        return None

    try:
        datetime_object = parser.parse(date_value, fuzzy=True)
        datetime_format = "%Y-%m-%dT%H:%M:%S"
        return datetime_object.strftime(datetime_format)
    except (ParserError, TypeError):
        logger.warning("Could not parse date: %s", date_value)
        return None
```

tests/search_gov_spiders/test_i14y_helper.py
```diff
@@ -1,70 +1,116 @@
...
+    sampleDates = {
+        "5/30/2024 7:24:49 AM": "2024-05-30T07:24:49",
+        "Wed, 02 Nov 1998": "1998-11-02T00:00:00",
+        "2025-03-14 00:00:00.0": "2025-03-14T00:00:00",
+        "2024-04-08 18:17:47-04:00": "2024-04-08T18:17:47",
+        "Thursday, August 10, 2023": "2023-08-10T00:00:00",
+        "January 8, 2013 10:05:30 AM EST": "2013-01-08T10:05:30",
+        "jibberish": None,
+        "2025-02-16T04:18:11.491+00:00": "2025-02-16T04:18:11",
+        "2024-02-22T00:00:00": "2024-02-22T00:00:00",
```
selfdanielj, line 23: I guess i still question if the timezones are being respected... like what timezone does Elasticsearch think these date strings are in?

IsabelLaurenceau: Not every date we get has timezones which is why I removed it all together. I can't convert it if I don't know what it is so rather than converting some but not all than assuming it's all in the same timezone I think it's better to leave the data as the time that it came in.

selfdanielj (commented):
looks good to me! Nice work

search_gov_crawler/elasticsearch/i14y_helper.py
```diff
@@ -20,18 +26,39 @@
...
+        str: Either the date in isoformat (YYYY-MM-DD'T'HH:MM:SS), or None
     """
-    return article_date or None
+    datetime_format = "%Y-%m-%dT%H:%M:%S"
+
+    if date_value is None or date_value == "":
+        return None
+
+    if isinstance(date_value, datetime):
+        return date_value.strftime(datetime_format)
```
igoristic, line 52: I don't think we should introduce or try to mimic ISO standard format. The datetime class has it builtin, so it would look something like this:
```suggestion
        return f"{date_value.isoformat()}Z"
```

search_gov_crawler/elasticsearch/i14y_helper.py
```diff
@@ -20,18 +26,39 @@
...
+
+    try:
+        datetime_object = parser.parse(date_value, fuzzy=True)
```
igoristic, line 55: We should ignore timezone or it will fail on dates like: `January 8, 2013 10:05:30 AM EST`
```suggestion
        datetime_object = parser.parse(date_value, ignoretz=True, fuzzy=True)
```
I think it's safer to assume that all timestamps are in UTC, or at least treat them as such. In some edge cases the worst that would happen is that we would be off by some hours (or one day)

search_gov_crawler/elasticsearch/i14y_helper.py
```diff
@@ -20,18 +26,39 @@
...
+        return datetime_object.strftime(datetime_format)
```
igoristic, line 56: Same suggestion as above: https://github.com/GSA-TTS/searchgov-spider/pull/143/files#r2015778526

igoristic (approved):
Looks good! 🏆 

Left a few comments/concerns.

Also, (in the future) I think we should do formatting/linting in a separate pull-requests. Makes it hard to review, or maybe I'm just dyslexia 🙃

search_gov_crawler/elasticsearch/i14y_helper.py
selfdanielj: if you pass a naive date string to elasticsearch, what timezone does it think its in?

igoristic: It sees it as UTC, and it is convention to store timestamps as UTC. I guess in this case, if we are supporting timezones, we should be sure to apply the offset to get the clean/pure UTC, something like `datetime_object.astimezone(timezone.utc)`

PR #152, opened 2025-03-28
SRCH-5906: Optimize PDF text extraction

igoristic:
## Summary
The most expensive function (memory wise) in pypdf is `extract_text`. We were calling this method twice for every page. This limits it to just one call per page, while maintaining original functionality

### Checklist

#### Functionality Checks

#### Process Checks


search_gov_crawler/elasticsearch/convert_pdf_i14y.py
```diff
@@ -47,8 +48,8 @@ def get_links_set(reader: PdfReader):
     ank = "/A"
     links = {}
 
-    for page in reader.pages:
-        text = page.extract_text()
+    for page_item in pages:
+        text, page = page_item
```
selfdanielj, line 52: Did you find that it was better to save the text twice for some reason? like you have the full text of all pages, why not just send that through the regex to get the links instead of storing the full text and the paginated text?

search_gov_crawler/elasticsearch/convert_pdf_i14y.py
```diff
@@ -32,7 +33,7 @@ def add_title_and_filename(key: str, title_key: str, doc: dict):
     doc[key] = f"{doc[title_key]} {doc['basename']}.{doc['extension']} {doc[key]}"
 
 
-def get_links_set(reader: PdfReader):
+def get_links_set(pages: list[Tuple[str, PageObject]]):
```
selfdanielj, line 36: In prod and in my testing i see a lot of `KeyError: '/A'` errors... seems like if there are no anchors then there are no links there so we don't need to raise a KeyError, right?

search_gov_crawler/elasticsearch/convert_pdf_i14y.py
```diff
@@ -47,8 +48,8 @@ def get_links_set(reader: PdfReader):
     ank = "/A"
     links = {}
```
selfdanielj, line 49: you call this function get_links_set but you use a dict with they key and value the same, why not just use a set datatype?

search_gov_crawler/elasticsearch/convert_pdf_i14y.py
```diff
@@ -142,9 +144,12 @@ def get_pdf_text(reader: PdfReader) -> str:
...
     """
     text = ""
+    pages = []
     for page in reader.pages:
-        text += page.extract_text() + " "
-    return text
+        page_text = page.extract_text()
+        text += page_text + " "
+        pages.append((page_text, page))
+    return (text, pages)
```
selfdanielj, line 152: related to a comment above, i question if its worthwhile to store both the full text and the paginated text instead of just using the full text to extract links.

search_gov_crawler/elasticsearch/convert_pdf_i14y.py
igoristic: It's not saving anything twice. I need to keep reference to the `PageObject` to be able to do `page.get_object()` in the links function. Yes, I can technically do it all in one loop/function, but I don't like to mix different functionalities/logic in **one** function. I think it's cleaner to have `get_links_set` and `get_pdf_text` separately. Also, nothing is really being "stored" we're simply passing reference and then discarding it, like in normal functional programming

search_gov_crawler/elasticsearch/convert_pdf_i14y.py
igoristic: Good catch! I'm checking if the key exists now and wrapped it in try/catch just incase

search_gov_crawler/elasticsearch/convert_pdf_i14y.py
igoristic: I changed it so only the links are the keys now. I think it's more of a preference though. I find that dict/maps are more universally intuitive since they are used for more use cases

search_gov_crawler/elasticsearch/convert_pdf_i14y.py
igoristic: I need to also pass the reference to the page, so I can do `page.get_object()` in the links

search_gov_crawler/elasticsearch/convert_pdf_i14y.py
selfdanielj: Don't change it, just want to say my piece... 

To me, when there is already a datatype that exists and you are using another datatype in a way so that you can get the functionality of the original data type, seems like it would be more intuitive to just use the data type designed for that use.

search_gov_crawler/elasticsearch/convert_pdf_i14y.py
selfdanielj: I was thinking that `text` which you build from all the extracted `page_text` values and all the `page_text` values that you return in the tuple are stored separately... but are essentially the same (minus some extra spaces you append).  I guess I actually don't know if references are used in this case or not.

search_gov_crawler/elasticsearch/convert_pdf_i14y.py
selfdanielj: I just re-ran the test domain that had the KeyError which is training.fema.gov and while I don't get that I do get this one now... not sure if this is an outlier or not... maybe the link value you extract needs to be cast as `str()` or otherwise decoded?
```

"ERROR", "message": "Failed to add item to Elasticsearch batch", "exc_info": "Traceback (most recent call last):
  File \"/Users/dself/Projects/searchgov-spider/search_gov_crawler/search_gov_spiders/pipelines.py\", line 93, in _process_es_item
      self._get_elasticsearch_client().add_to_batch(
	    File \"/Users/dself/Projects/searchgov-spider/search_gov_crawler/elasticsearch/es_batch_upload.py\", line 49, in add_to_batch
		    doc = convert_pdf(response_bytes=response_bytes, url=url, response_language=response_language)
			          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
					    File \"/Users/dself/Projects/searchgov-spider/search_gov_crawler/elasticsearch/convert_pdf_i14y.py\", line 135, in convert_pdf
						    i14y_doc[content_key] = f\"{i14y_doc[content_key]} {' '.join(all_links) if len(all_links) > 0 else ''}\"\n{i14y_doc[content_key]} {' '.join(all_links) if len(all_links) > 0 else ''}\"
							                                                       ^^^^^^^^^^^^^^^^^^^\nTypeError: sequence item 0: expected str instance, ByteStringObject found" 
```

igoristic: Nice catch!
Looks like pypdf might return bytes instead of strings for some pdf objects

selfdanielj (approved):
Looks good to me! Nice work.

PR #163, opened 2025-04-03
SRCH-5903 human schedule view

IsabelLaurenceau:
## Summary
Creates markdown files of schedules. 

### Testing 
Includes two functionalities: 

1) Create all schedules by running: 
```bash
cd search_gov_crawler/domains
python readschedule.py
```

2) Create individual schedules by running: 
```
cd search_gov_crawler/domains
python readschedule.py crawl-sites-development.json
python readschedule.py crawl-sites-staging.json
python readschedule.py crawl-sites-production.json   
```

The current markdowns are already included in this PR however you can delete them and run the instructions above to see functionality. 

### Checklist

#### Functionality Checks

#### Process Checks

### DIFF
--- a/search_gov_crawler/domains/readschedule.py
+++ b/search_gov_crawler/domains/readschedule.py
@@ -111,5 +111,5 @@ if __name__ == "__main__":
         main(sys.argv[1])
     else:
         schedules = ["crawl-sites-development.json", "crawl-sites-staging.json", "crawl-sites-production.json"]
-        for schedule in schedules:
-            main(schedule)
+        for json_schedule in schedules:
+            main(json_schedule)
--- a/search_gov_crawler/domains/readschedule.py
+++ b/search_gov_crawler/domains/readschedule.py
@@ -85,9 +85,14 @@ def create_sorted_markdown(unsorted_schedule, file_path):
 
 
 def create_markdown_schedule_file(file_path):
-    file_name = file_path.split(".")[0] + ".md"
+    env = file_path.split(".")[0]
+    file_name = env + ".md"
+    title = env.split("-")[-1]
+    header = "# " + title.capitalize() + " Schedule\n"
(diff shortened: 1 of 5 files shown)


search_gov_crawler/domains/readschedule.py
```diff
@@ -0,0 +1,115 @@
...
+            case "SUN":
+                unsorted_schedule[6].append({"name": name, "time": time})
+    return unsorted_schedule
+
+
+def create_markdown(day, data, file_name):
+    if not data:
+        return ""
+
+    md_table = f"""## {day} \n|Domain|Time (UST)|\n|---|---|\n"""
```
selfdanielj, line 62: `UST` should be `UTC` here right?

IsabelLaurenceau: yup

selfdanielj (approved):
Nice work! It works for me to generate the schedules. As an initial step I think we can go with this.

If we are going to make this more of the process, for instance generating updated calendars each time we change the jsonnet files, especially if its automated or checked in the CI process, I would like to see:
* doc strings for the functions explaining what they do
* unit tests
* a more straightforward flow... as is we follow functions that call functions that call functions and for me it gets hard to follow.  I think it would be easier to understand whats going on if more of the logic was in main, for instance something like:
```python
    data = read_json_file(file_path)
    if data is not None:
        unsorted_schedule = put_into_schedule_format(data)
        toc, header = create_markdown_schedule_file(file_path) # changed to not write file just return toc, header

        md_tables = []
        days_names = ["Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday", "Sunday"]
        for i, schedule in enumerate(unsorted_schedule):
            day = days_names[i]
            sorted_by_time = sorted(schedule, key=lambda x: x["time"])
            md_tables.append(create_markdown(day, sorted_by_time, file_path)) # return the md_table instead of calling write schedule

        write_schedule(toc, file_path) # or just call create_markdown_schedule_file as-is here
        write_schedule(header, file_path)
        write_schedule("".join(md_tables), file_path)
```

PR #172, opened 2025-04-09
SRCH-5469-re-add-spidermon

IsabelLaurenceau:
## Summary
Adding back spidermon monitoring and configuring it with AWS SES (simple email service). 

### To set up in AWS for other environments (already set in dev) : 
- Give **dgsearch-images-manager** role `ses:SendRawEmail` permissions 
- verify the identity of **tts-search-devs@gsa.gov** email on ses platform https://us-east-1.console.aws.amazon.com/ses/home?region=us-east-1#/identities

### Testing: 
- connect to ec2 instance using session manager: https://us-west-1.console.aws.amazon.com/ec2/home?region=us-west-1#ConnectToInstance:instanceId=i-0c18717ce265ce348 and follow instructions below 

**In the ec2:** 
- sudo su - search
- cd $PYTHONPATH 
- source venv/bin/activate

- make sure the new necessary env var are there (SEARCH_AWS_ACCESS_KEY_ID,   SEARCH_AWS_SECRET_ACCESS_KEY,   SPIDERMON_EMAIL_SENDER, SPIDERMON_EMAIL_TO)

- cd search_gov_crawler

This should not send any email (i.e it should work fine) 
- scrapy crawl domain_spider -a allowed_domains=quotes.toscrape.com -a start_urls=https://quotes.toscrape.com -a output_target=csv -a depth_limit=2

This should send an email (i.e it triggers one of the monitors) 
- scrapy crawl domain_spider -a allowed_domains=quotes.toscrape.com/js/ -a start_urls=https://quotes.toscrape.com/js/ -a output_target=csv -a depth_limit=2

### Checklist

#### Functionality Checks

#### Process Checks


search_gov_crawler/search_gov_spiders/settings.py
```diff
@@ -109,10 +109,10 @@
 date_time = spider_start.isoformat()
 body_html_template = Path(__file__).parent / "actions" / "results.jinja"
 
-SPIDERMON_ENABLED = os.environ.get("SPIDERMON_ENABLED", "False")
-SPIDERMON_MIN_ITEMS = 1000
-SPIDERMON_TIME_INTERVAL = 1  # time is in seconds
-SPIDERMON_ITEM_COUNT_INCREASE = 100
+SPIDERMON_ENABLED = os.environ.get("SPIDERMON_ENABLED", "True")
```
selfdanielj, line 112: What about changing this to SPIDER_SPIDERMON_ENABLED (to match our naming of other env vars) and putting it in parameter store (or creating a story for that)?  I am thinking we don't want this on in all environments?

search_gov_crawler/search_gov_spiders/settings.py
```diff
@@ -121,15 +121,12 @@
...
-SPIDERMON_SMTP_USER = os.environ.get("SPIDERMON_SMTP_USER")
-SPIDERMON_SMTP_PASSWORD = os.environ.get("SPIDERMON_SMTP_PASSWORD")
-SPIDERMON_SMTP_ENFORCE_SSL = False
-SPIDERMON_SMTP_ENFORCE_TLS = True
-
+SPIDERMON_EMAIL_SENDER = "search@support.digitalgov.gov"
+SPIDERMON_EMAIL_TO = "tts-search-devs@gsa.gov"
+SPIDERMON_AWS_ACCESS_KEY_ID = os.environ.get("SEARCH_AWS_ACCESS_KEY_ID")
+SPIDERMON_AWS_SECRET_ACCESS_KEY = os.environ.get("SEARCH_AWS_SECRET_ACCESS_KEY")
+SPIDERMON_AWS_REGION_NAME = "us-east-1"
```
selfdanielj, line 128: Does email send work locally?  If not, i wonder if there is a way to turn if off when running locally.  We could disable spidermon locally, perhaps have the default for SPIDERMON_ENABLED be "False" and enable it where we want it.  I'm not sure if there is a benefit of getting the report and not sending the email but if so it would be nice to just disable the email send.  Could be a future improvement as well i guess.

search_gov_crawler/search_gov_spiders/monitors.py
selfdanielj, line 1: i would love some documentation here or maybe in a readme explaining a bit about our usage of spidermon... not sure exactly what level of detail makes sense.

search_gov_crawler/search_gov_spiders/settings.py
IsabelLaurenceau: Why wouldn't we want it in all env? Wouldn't we want to know if something we are testing is not working in dev and stage?

selfdanielj: hmm yeah you're right we probably would want to know.  I guess i'm just worried about getting too many emails but i don't really know how many we'll get.  is there a way to distinguish between emails sent from dev/staging/prod like maybe adding something to the subject?

search_gov_crawler/search_gov_spiders/settings.py
IsabelLaurenceau: If you have SEARCH_AWS_ACCESS_KEY_ID and SEARCH_AWS_SECRET_ACCESS_KEY on your local as environment vars it works

search_gov_crawler/search_gov_spiders/monitors.py
IsabelLaurenceau: I can add that. I used to have it in confluence but I just checked and it's actually outdated anyway

search_gov_crawler/search_gov_spiders/settings.py
selfdanielj: ok good to know... if our local testing could also be sending emails to the same address then i would really want a way to easily tell which environment the report is about.

selfdanielj:
"Give dgsearch-images-manager role ses:SendRawEmail permissions" is this something that should be added in terraform? or did Luke do this already in a separate TF branch?

IsabelLaurenceau:
> "Give dgsearch-images-manager role ses:SendRawEmail permissions" is this something that should be added in terraform? or did Luke do this already in a separate TF branch?

I think he did it

search_gov_crawler/search_gov_spiders/settings.py
IsabelLaurenceau: Added the environment name to the email subject line

selfdanielj (commented):
I was not able to get it to work locally due to AWS permissinos but changes seem ok except for when i was looking at the messages from dev in the inbox, I couldn't tell which domain they were for.  Is there a way to include that info in the report?

IsabelLaurenceau:
> I was not able to get it to work locally due to AWS permissions but changes seem ok except for when i was looking at the messages from dev in the inbox, I couldn't tell which domain they were for. Is there a way to include that info in the report?

Updated the code. You can check the inbox to see that the domain is now in the header

selfdanielj:
> > I was not able to get it to work locally due to AWS permissions but changes seem ok except for when i was looking at the messages from dev in the inbox, I couldn't tell which domain they were for. Is there a way to include that info in the report?
> 
> Updated the code. You can check the inbox to see that the domain is now in the header

Looks good!  And I see how we can hopefully expand if it turns out wee need more info.

selfdanielj (approved):
Nice work!

PR #183, opened 2025-04-21
SRCH-5961 spider schedule to redis

selfdanielj:
## Summary
- Moves scheduler state into redis
  - Uses three keys for spider in db1: one for job state, one for next run time, one for pending jobs
  - Jobs are added to the pending queue when they are submitted to run (regardless of if they actually run right then)
  - Jobs are removed from the pending queue when they fail or succeed
  - Any jobs in the pending queue when the scheduler starts stay at the top of the queue and are immediately run first
  - Supports changes to existing jobs on deploy
- Extended APScheduler scheduler and jobstore classes to provide extra functionality (these are in search_gov_cralwer/scheduling).
- Updated scrapy_scheduler to perform necessary actions on restart.
  - Moved from BlockingScheduler to BackgroundScheduler so that I could perform these actions.  When using BlockingScheduler you cannot adjust the scheduler once you start it so I would not be able to mark jobs in the pending job store as needed to rerun immediately.
- Did not change benchmark -- I thought this should remain as an in-memory scheduler.
- Dependency Updates
  - Added redis library to dependencies
  - Included apscheduler in dependabot file
- Connectivity
  - Adds redis container to localdev docker compose file
  - Grabs existing env var values from parameter store for connectivity in higher environments
 

### Testing
- Run `docker compose up` to start the local dev containers (or just `docker compose up redis` if you are already running the others)
  - Ensure you can connect to the local redis instance
- Start scrapy scheduler -- `python search_gov_crawler/scrapy_scheduler.py`
  - Ensure you can see two keys in db1 on the local redis instance with data for all the jobs in the scheduler
- You can either, wait for a job or jobs to run or stop the scheduler (ctrl-C or whatever), adjust a few of the jobs schedules so that they run very soon.  Restart the scheduler and wait for jobs to start.
  - Be sure to check your SPIDER_SCRAPY_MAX_WORKERS variable.  Maybe set it to 1 so you can get a better view of jobs in the queue that are not running.
- Verify some key functionality:
  - When jobs are at their scheduled run time they are added to the pending queue
  - When jobs complete successfully they are removed from the pending queue
  - When jobs complete with an error they are removed from the pending queue
  - The scheduler can be stopped and the state remains in redis
  - The scheduler can be restarted and jobs in the pending queue are immediately triggered for rerun (and are prefixed with "Rerun::"
  - You can add, edit, or remove jobs from the jsonnet during a restart and the changes will be applied to redis.
  - If a job is removed from the jsonnet while it has an entry in the pending queue, that job is removed from the pending queue.
- Whatever else you want to test

### Checklist

#### Functionality Checks

#### Process Checks


igoristic (approved):
Looks good 🎯  Great job!

PR #189, opened 2025-04-24
SRCH-5940 Delete by Domain

selfdanielj:
## Summary
- Script to allow us to delete data from the spider index for a given domain
  - Used print and didn't add unit tests since this is not prod code... let me know if you think we need logging or tests.
- Uses SearchGovElasticsearch to tie index existing creds instead of recreating all that.
  - Added properties to the class so i don't have to access private attributes
  - Added tests to cover these
- Forces user to use flag to apply deletes 
  - I thought about adding some more validation on controls to what the user can do but wasn't sure what was appropriate or needed since its just "us" doing it.
- Moved convert_libsonnet into the new scripts folder.
  - I was on the fence about the name of the folder since we now have "scripts" and "cicd-scripts". I think I'd like to have one folder for all of these at some point.

### Testing
- Load some documents into your local elasticsearch
- Run the command without the apply flag: `python scripts/delete_by_domain.py example.com`
  - Verify output and confirm no deletion occurs
- Run the command with the apply flag: `python scripts/delete_by_domain.py example.com --apply`
  - Verify output and confirm deletions.
- Try whatever else you can think of to break it?

### Checklist

#### Functionality Checks

#### Process Checks


igoristic (approved):
Looks good 👍

PR #199, opened 2025-04-29
SRCH-5988 update spidermon settings

selfdanielj:
## Summary
- Goal here is to reduce the emails and logging from spidermon.  Its so much that its not helpful.

Changes
- Removed the periodic monitor check for change in item count.
- Adjusted the periodic monitor check interval from once a minute to once an hour
- Added a status check on end of spider that alerts if it stops for reasons other than "Finished"
- Adjusted other settings so that we don't get as many alerts:
  - Reduced minimum items check to 1.  Some sites are small.  If we get 0 i would like to know otherwise not.
  - Raised max run time to 48 hours since that seems to be the threshold that indicates a problem.
  - Raised max codes count to 50 since some of these sites are large and may have a large number of errors.

# Testing
- Run a job, ensure no configuration errors
- You should see spidermon logging at the end of the job for the close monitor and if it runs for longer than an hour you should see more logging for the periodic monitor.
- If there are failures you should see an email get sent

### Checklist

#### Functionality Checks

#### Process Checks

### DIFF
--- a/search_gov_crawler/search_gov_spiders/monitors.py
+++ b/search_gov_crawler/search_gov_spiders/monitors.py
@@ -1,42 +1,48 @@
 from pathlib import Path
+from typing import ClassVar
+
 from spidermon import MonitorSuite
 from spidermon.contrib.actions.email.ses import SendSESEmail
 from spidermon.contrib.actions.reports.files import CreateFileReport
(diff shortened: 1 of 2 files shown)


igoristic (approved):
Looks good 🥇

PR #205, opened 2025-05-01
SRCH-5953 Log Errors When Occurs On Starting Urls

selfdanielj:
## Summary
The goal of this story is to log an `ERROR` level message when a spider encounters an error with one or more of the start URLs.  This only happens now if there is a DNS, timeout, or other connection error.  If there is a 403 or the start URL redirects out of the domain, we sometimes get an INFO message but it does not get into the cloudwatch filters that show log messages where `levelname == ERROR`.
- Originally considered sub-classing the built-in `HttpErrorMiddleware`
- But since we already have a `SearchGovSpidersSpiderMiddleware` that is not really doing anything, I just added a check in `process_spider_exception` and logged the error there.  Seemed cleaner not to mess with the built-in middleware if I didn't have to.
- Also needed to update the `SearchGovSpidersOffsiteMiddleware` to let us know when the starting url generates an offsite request.  Its been hard to find one of these working locally (since most of the AF domain examples give 403s locally) but if you set the starting url and allowed domain to different values, it works.
- Updated tests

### Testing
- Run a working crawl (quotes.toscrape.com or whatever) and confirm no logging issues.
- Running a crawl locally for `{ https://www.rs.af.mil/ | rs.af.mil }` returns 0 results due to a 403.  With this change, there should be a log message like  `Error occured while accessing start url:...` and info about the error
- You can try others that return a DNSError like `{ https://www.hancockfield.ang.af.mil/ | hancockfield.ang.af.mil }` but these were already logging an error.
- To mimic an Offsite Error run a scrape where the starting URL is not in the allowed domains like `{ https://quotes.toscrape.com | example.com }`

### Checklist

#### Functionality Checks

#### Process Checks

### DIFF
--- a/search_gov_crawler/search_gov_spiders/middlewares.py
+++ b/search_gov_crawler/search_gov_spiders/middlewares.py
@@ -1,7 +1,8 @@
-# Define here the models for your spider middleware
-#
-# See documentation in:
-# https://docs.scrapy.org/en/latest/topics/spider-middleware.html
+"""Define here the models for your spider middleware
+
+See documentation in:
+https://docs.scrapy.org/en/latest/topics/spider-middleware.html
+"""
 
 import re
 import warnings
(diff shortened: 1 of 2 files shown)


igoristic (approved):
Looks good 🚀

PR #210, opened 2025-05-06
Srch 5932 cannot access stopwords bug

selfdanielj:
## Summary
- Bug is we are seeing logs of errors about missing stop words files.  This is because for every doc we are trying to use the stopwords files from nltk and nltk does not support all the languages we support.
- Took two actions to fix this:
  - For HTML, since `summarize_text` where the error occurs is not the primary method for generating the `description` or `tags` fields, I moved it into a conditional so that it only runs when if it is needed.  Many of the error messages were appearing for cases when newspaper4k or `convert_html_scrapy` already generated values and the `summarize_text` values would be discarded anyway.  I did not do the same for PDFs since `summarize_text` is a primary data-generation method there.
  - Clarified the reasons why we return None from `summarize_text` and added a try/except block around getting the language from nltk.  Now, if we try to get a stopwords file that doesn't exist it logs a WARNING.
  - I added `url` as a argument so we can have some visibility in the logs as to which URL caused the error since this is not a spider logger we don't have the added spider stats.  I felt a little weird adding the arg just for logging but maybe not that bad.
  - Added a few tests.

### Testing
It was actually hard to find an example that logged an warning after these changes (which i think is a good thing and a result of moving `summarize_text` into a conditional.
I did find one:

Run 
`scrapy crawl domain_spider -a allowed_domains=uscis.gov -a start_urls="https://www.uscis.gov/humanitarian/humanitarian-or-significant-public-benefit-parole-for-aliens-outside-the-united-states/the-haitian-family-reunification-parole-hfrp-program-haitian-creole" -a output_target=elasticsearch -a depth_limit=1`

It takes a few minutes to finish but if you look back through the logs you will no OSError exceptions, but instead see:
`{"asctime": "2025-05-06 10:50:11,008", "name": "search_gov_crawler.elasticsearch.i14y_helper", "levelname": "WARNING", "message": "Unsupported Language. Error when parsing https://www.uscis.gov/sites/default/files/document/guides/HFRP-How-to-Check-If-Your-IV-is-Available-Haitian-Creole.pdf Missing Stopwords File: No such file or directory: '/Users/dself/nltk_data/corpora/stopwords/croatian'"}`

### Checklist

#### Functionality Checks

#### Process Checks

### DIFF
--- a/search_gov_crawler/elasticsearch/convert_html_i14y.py
+++ b/search_gov_crawler/elasticsearch/convert_html_i14y.py
@@ -1,18 +1,18 @@
 import newspaper
-from search_gov_crawler.elasticsearch.parse_html_scrapy import convert_html_scrapy
-from search_gov_crawler.search_gov_spiders.helpers import content
-from search_gov_crawler.search_gov_spiders.helpers import encoding
+
 from search_gov_crawler.elasticsearch.i14y_helper import (
     ALLOWED_LANGUAGE_CODE,
-    parse_date_safely,
-    get_url_path,
-    get_base_extension,
     current_utc_iso,
-    generate_url_sha256,
-    get_domain_name,
     detect_lang,
+    generate_url_sha256,
+    get_base_extension,
+    get_domain_name,
+    get_url_path,
+    parse_date_safely,
     summarize_text,
 )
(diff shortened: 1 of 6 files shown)


igoristic (approved):
Looks good :trophy:

PR #184, opened 2025-04-21
SRCH-5886: Check sitemaps for updates and trigger indexing if there's new content

igoristic:
#### Functionality Checks

- [x] You have merged the latest changes from the target branch (usually `main`) into your branch.

- [x] Your primary commit message is of the format **SRCH-#### \<description\>** matching the associated Jira ticket.

- [x] PR title is either of the format **SRCH-#### \<description\>** matching the associated Jira ticket (i.e. "SRCH-123 implement feature X"), or **Release - SRCH-####, SRCH-####, SRCH-####** matching the Jira ticket numbers in the release.

- [x] Automated checks pass. If Code Climate checks do not pass, explain reason for failures:

#### Process Checks

- [x] You have specified at least one "Reviewer".

PR #220, opened 2025-05-16
Staging Release - SRCH-5886 SRCH-6014

selfdanielj:
SRCH-5886: Check sitemaps for updates and trigger indexing if there's new content (#184)
SRCH-6014 add deny path (#219)


igoristic (approved):
Looks good :shipit:

PR #224, opened 2025-05-20
Production Release SRCH-5886 SRCH-6014

selfdanielj:
SRCH-6014 add deny path (#219) 
SRCH-5886: Check sitemaps for updates and trigger indexing if there's new content (#184)

@igoristic


jimkreft (approved):
Approved

PR #230, opened 2025-05-29
SRCH-6029: Batch processing new sitemap URLs

igoristic:
There might be a limit as to how many starting URLs we can have with scrapy. Adding batches with delay does not overload the resources by trying to do all the starting URLs at once

#### Functionality Checks

#### Process Checks

### DIFF
--- a/search_gov_crawler/search_gov_spiders/sitemaps/sitemap_monitor.py
+++ b/search_gov_crawler/search_gov_spiders/sitemaps/sitemap_monitor.py
@@ -287,25 +287,41 @@ class SitemapMonitor:
                     log.info("\n".join(new_urls_msg_lines))
 
                     record = self.records_map[sitemap_url]
-                    spider_args = {
-                        "allow_query_string": record.allow_query_string,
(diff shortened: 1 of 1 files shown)


selfdanielj (changes requested):
I get an error when i remove 30 urls from an existing spider_sitemaps file.

```
    batch = new_urls[start:end]
            ~~~~~~~~^^^^^^^^^^^
TypeError: 'set' object is not subscriptable
```

Generally I think the logic will work as-is if you cast it to a list (`batch = list(new_urls)[start:end]` or something) but did you consider using https://docs.python.org/3/library/itertools.html#itertools.batched?? might be a little cleaner, not having to worry about the math or checking for the end of the batches

igoristic:
@selfdanielj 
> I get an error when i remove 30 urls from an existing spider_sitemaps file...

Nice catch! For some reason I can't reproduce that same error. But, think I covered that scenario with the new code

selfdanielj (approved):
👍

PR #235, opened 2025-06-03
SRCH-6044: Added sitemap extension check

igoristic:
#### Functionality Checks

- [x] You have merged the latest changes from the target branch (usually `main`) into your branch.

- [x] Your primary commit message is of the format **SRCH-#### \<description\>** matching the associated Jira ticket.

- [x] PR title is either of the format **SRCH-#### \<description\>** matching the associated Jira ticket (i.e. "SRCH-123 implement feature X"), or **Release - SRCH-####, SRCH-####, SRCH-####** matching the Jira ticket numbers in the release.

- [x] Automated checks pass. If Code Climate checks do not pass, explain reason for failures:

#### Process Checks

- [x] You have specified at least one "Reviewer".

PR #250, opened 2025-06-18
Production Release - SRCH-6044

selfdanielj:
SRCH-6044: Added sitemap extension check (#235)
Update to dependabot ignores (#246)


igoristic (approved):
Looks good ✅

PR #253, opened 2025-06-20
SRCH-6051 DAP Extractor Process

selfdanielj:
## Summary
- New standalone process to get visits data from the DAP API and store them in Redis.
  - Gets data for a parameterized number of days
  - Stores in redis by domain
  - Removes data older than a parameterized number of days
  - Runs on a schedule (probably daily)
- Design Decisions
  - Created new entrypoint file `search_gov_crawler/dap_extractor.py`
  - Other code in new `search_gov_crawler/dap` module
  - Used `APScheduler` to schedule instead of a cron job since that's what we are doing with scrapy scheduler and we can start it the same way as scrapy scheduler, sitemap, etc...
  - Used different DB in redis to store these keys (db=2) to provide separation from 
- Updated cicd scripts
  - Wrote to same log file (maybe we need a new story to split into three?)
  - Added new env variables to fetch script
    - Related PR in tf repo: GSA/searchgov-tf#178
  - Added kill helper but its the same as the sitemap kill script, wanted to make this generic but thought it was too out of scope. 
- Updated README and other docs
- Added test coverage
- Also added a function to get average daily visits for a domain but that is not used yet, just wanted to show how we can do it.  Will probably make updates to it as we need it.

## Testing
- Set env variables locally
- Run
  - On schedule from env vars `python search_gov_crawler/dap_extractor.py`
  - On schedule overriding parameters `python search_gov_crawler/dap_extractor.py --days-back 30 --max-age 100`
  - On demand `python search_gov_crawler/dap_extractor.py --run-now`
- After each run check redis db=2 to see what was loaded.

### Checklist

#### Functionality Checks

#### Process Checks


search_gov_crawler/dap/connect.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+    dap_api_base_url, dap_api_key = get_dap_api_configs()
+
+    url = (
+        f"{dap_api_base_url}/reports/site/data?"
+        f"after={query_date}&before={query_date}&"
+        f"limit={query_limit}&page={query_page}"
+    )
+
+    response = requests.get(url, headers={"x-api-key": dap_api_key, "Accept": "application/json"}, timeout=30)
```
igoristic, line 39: I don't think we should handle the encoding and building of URL params ourselves. We can use python's built in options.
```suggestion
    base_url, api_key = get_dap_api_configs()
    endpoint = f"{base_url}/reports/site/data"

    params = {
        "after": query_date,
        "before": query_date,
        "limit": query_limit,
        "page": query_page,
    }

    headers = {
        "x-api-key": api_key,
        "Accept": "application/json",
    }

    response = _SESSION.get(endpoint, params=params, headers=headers, timeout=30)
```

Also, I added sessions here because I know they are faster for retry/resolve if we are doing header type authentication (via API key). You'll just need to add/hoist: `_SESSION = requests.Session()` somewhere on top outside of the function

selfdanielj: yeah good call, i guess i didnt go this route initially because we control the input but we should probably be doing it all the time.  I'll make these changes.

search_gov_crawler/dap/datastore.py
```diff
@@ -0,0 +1,64 @@
...
+
+def age_off_dap_records(redis: Redis, days_back: int) -> int:
+    """
+    Removes DAP records older than `days_back` days.
+    """
+
+    dap_records_aged_off = 0
+    age_off_date = datetime.now(UTC) - timedelta(days=days_back)
+    min_score = int(age_off_date.strftime("%Y%m%d"))
+    keys = redis.keys("dap_visits:*")
```
igoristic, line 29: redis.keys can be an expensive (and blocking) operation (if we have lots of keys). Maybe you can do a scan instead?
```suggestion
    keys = list(redis.scan_iter("dap_visits:*"))
```

search_gov_crawler/dap/datastore.py
```diff
@@ -0,0 +1,64 @@
...
+    min_score = int((datetime.now(UTC) - timedelta(days=days_back)).strftime("%Y%m%d"))
+    max_score = int(datetime.now(UTC).strftime("%Y%m%d"))
+
+    domain_keys = list(redis.keys(f"{key_prefix}:*.{domain}"))
+    domain_keys.extend(redis.keys(f"{key_prefix}:{domain}"))
+
+    for domain_key in domain_keys:
+        visits = redis.zrangebyscore(name=domain_key, min=min_score, max=max_score)
+        avg_daily_visits = sum(int(visit) for visit in visits) / days_back if visits else 0
+        domain_name = str(domain_key).removeprefix(key_prefix)
```
igoristic, line 54: NIT: Not sure how this is working. Doesn't Redis return bytes? I'm use to doing decoding it first
```suggestion
    for raw_key in domain_keys:
        key = raw_key.decode() if isinstance(raw_key, bytes) else raw_key
        visits = redis_client.zrangebyscore(key, min=min_score, max=max_score)
        avg_daily_visits = sum(int(visit) for visit in visits) / days_back if visits else 0
        domain_name = key[len(key_prefix):]
```

search_gov_crawler/dap/datastore.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+
+    dap_records_aged_off = 0
+    age_off_date = datetime.now(UTC) - timedelta(days=days_back)
+    min_score = int(age_off_date.strftime("%Y%m%d"))
+    keys = redis.keys("dap_visits:*")
+
+    log.info("Checking %d keys for records oder than %s", len(keys), age_off_date.strftime("%Y-%m-%d"))
+    for key in keys:
+        dap_records_aged_off += redis.zremrangebyscore(name=key, min="-inf", max=min_score)
```
igoristic, line 33: I think in cases like this you should use a redis pipeline to avoid round trips, eg:
```suggestion
    pipe = redis_client.pipeline()
    for key in keys:
        pipe.zremrangebyscore(key, "-inf", min_score)
    results = pipe.execute()
    dap_records_aged_off = sum(results)
```

igoristic (dismissed):
Great job 💪  Some minor comments, but I don't think any of them are blockers. Also, I keep seeing this hardcoded `db=2` for redis should we have it as a global var somewhere (or maybe environment variable)?

PR #262, opened 2025-07-01
SRCH-6052 DAP Visits On Doc

selfdanielj:
## Summary
- Added functionality so that the `dap_domain_visits_count` field on ES documents produced by spider contain either a 0 or a seven day average of the visits data extracted from DAP.
  - Updated existing `get_avg_daily_visits_by_domain` to match logic discussed in the dev channel, use pipelines, and explicitly decode responses.
  - In spider `__init__`, added helper function `get_domain_visits` that retrieves all domain and sub-domain data for all values in the `allowed_domains`, e.g. for a spider crawl of `gsa.gov` this will contain DAP records for `gsa.gov`, `a.gsa.gov`, `b.gsa.gov`, etc...
  - Added default values of 0 to `dap_domain_visits_count` during document creation.
  - In the `add_to_batch` method of `SearchGovElasticsearch`, added a function `update_dap_visits_to_document` that updates the documents after doing a lookup to the data retrieved during `__init__` based on the `domain_name` value in the document.
- Also removed the functionality to disable PDFs, since I don't think we are planning on doing that any more.
- Updated a bunch of tests and added test coverage where is was easy to do so.

### Testing
- Run DAP extractor `python search_gov_crawler/dap_extractor.py --days-back 14 --max-age 28 --run-now`
- Delete local searchelastic index
- Create new local searchelastic index (if you want to easily search by domain_name you should do so by using the customized settings and mappings we have in the repo but its not required for this story.)
- Run a spider job for one of the domains captured in the DAP data and assess the results.
- I tested with fac.gov
```
scrapy crawl domain_spider \
    -a allowed_domains=fac.gov \
    -a start_urls="https://www.fac.gov" \
    -a output_target=elasticsearch \
    -a depth_limit=8
```
- Verify that documents loaded into elasticsearch match a query like `GET /searchgov-spider/_search
{ "query": { "term": { "domain_name": { "value": "fac.gov" } } } }` have expected values for `dap_domain_visits_count`
  - For `domain_name` values of `www.fac.gov` there should be a value matching the last seven days average in the dap_visits:fac.gov` redis key.
  - For `domain_name` values of 'support.fac.gov` there should be a 0 value because there is not a `dap_visits:support.fac.gov`

### Checklist

#### Functionality Checks

#### Process Checks

### DIFF
--- a/search_gov_crawler/dap/datastore.py
+++ b/search_gov_crawler/dap/datastore.py
@@ -41,24 +41,37 @@ def get_avg_daily_visits_by_domain(redis: Redis, domain: str, days_back: int) ->
     """
     key_prefix = "dap_visits:"
 
-    domain_avg_daily_visits = {}
-    min_score = int((datetime.now(UTC) - timedelta(days=days_back)).strftime("%Y%m%d"))
-    max_score = int(datetime.now(UTC).strftime("%Y%m%d"))
+    # Use yesterday as basis for these calculations to lessen the impact of DAP data not being available
+    base_datetime = datetime.now(UTC) - timedelta(days=1)
+    min_score = int((base_datetime - timedelta(days=days_back)).strftime("%Y%m%d"))
+    max_score = int(base_datetime.strftime("%Y%m%d"))
 
-    domain_keys = list(redis.keys(f"{key_prefix}:*.{domain}"))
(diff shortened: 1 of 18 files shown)


igoristic (approved):
Looks good 🚀

PR #265, opened 2025-07-09
SRCH-6054 Testing Search Scoring

selfdanielj:
## Summary
- We talked about trying a few things to [test query changes](https://docs.google.com/document/d/1pOx4FP6wZ9iOIL1WZM1fxXlHBmd9-jdMde1C5v-m8sQ/edit?tab=t.0#heading=h.vrg0i1i30ip2) 
- I was having trouble getting a complex query like the one the search app uses for search results to work inside Kibana
- So i wrote this script, the idea being we could sit with David and Ammie and run some searches and compare results.
- This uses the production search query which I adjusted the query to include the new scoring function AND to be a full government search
- So when you run this in prod, it should show you what the results will be if we make this change.
- I had to make the TYPE_CHECKING change since this script was causing a circular dependency due to my TypeVar that I added recently, but really that should have been there to begin with.
- Added rich as a testing dependency to help with output.

### Testing
- Run pip install to get the new dependencies
- Run a search using various options:
  - Search for "documents"
    - `python scripts/query_tester.py full-govt-search documents`
  - Same thing but only show five results instead of 10
    - `python scripts/query_tester.py full-govt-search documents --size 5`
  - Same thing but see page 2
    - `python scripts/query_tester.py full-govt-search documents --size 5 --page 2`
  - Search using a multi word query with page size 20 
    - `python scripts/query_tester.py full-govt-search social security --size 20`

### Checklist

#### Functionality Checks

#### Process Checks

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -27,6 +27,7 @@ updates:
           - "freezegun"
           - "pylint"
           - "pytest*"
+          - "rich"
     ignore:
       # see requirements.txt for more details
       - dependency-name: "lxml-html-clean"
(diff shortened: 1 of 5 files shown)


igoristic (approved):
Looks good 🏆

PR #285, opened 2025-07-30
Staging Release - SRCH-6120 SRCH-6129

selfdanielj:
SRCH-6120 New URL changes (#283) 
SRCH-6129 update metadata date parsing to reduce logging and to not prevent documents from loading (#282)
SRCH-6129 update regex to escape dash (#286)


igoristic (approved):
Looks good 🏆

PR #314, opened 2025-08-29
SRCH-6199 Increase logging level to avoid irrelevant error logging

selfdanielj:
## Summary
- Ive been seeing these ERROR messages in the logs: 
```json
{
  "levelname": "ERROR",
  "message": "Advanced encoding /SymbolSetEncoding not implemented yet",
  "name": "pypdf._cmap"
}
```
- They have been annoying me since they don't appear to actually break anything and sometimes there are a lot of them.   Also they don't say which PDF triggers the log message so its been hard to find an example to test on.
- Finally found one on a random github page and raised the log level for that logger.
- It doesn't seem to really affect the content, at least in this one example I found so I thought its ok to suppress the message.

### Testing
- Here is the file I found that triggers a log message
```
scrapy crawl domain_spider \
    -a allowed_domains=github.com \
    -a start_urls="https://github.com/UglyToad/PdfPig/files/10214328/ZapfDingbatsFontAddText.pdf" \
    -a output_target=elasticsearch \
    -a depth_limit=1 \
    -a prevent_follow=True
```
- Run in main and you will see an error message: `{"asctime": "2025-08-29 12:40:25,618", "name": "pypdf._cmap", "levelname": "ERROR", "message": "Advanced encoding /FontSpecific not implemented yet"}`
- Run in this branch and the error is not there
- You can check elasticsearch to see the doc. The URL ends up being different but since it was the last doc you loaded you should be able to use
```
GET /development-i14y-documents-searchgov/_search
{
  "query": {
    "match_all": {}
  },
  "sort": [
    {
      "updated_at": {
        "order": "desc"
      }
    }
  ],
  "size": 1
}
```

### Checklist
Please ensure you have addressed all concerns below before marking a PR "ready for review" or before requesting a re-review. If you cannot complete an item below, replace the checkbox with the ⚠️ `:warning:` emoji and explain why the step was not completed.

#### Functionality Checks

- [X] You have merged the latest changes from the target branch (usually `main`) into your branch.

- [X] Your primary commit message is of the format **SRCH-#### \<description\>** matching the associated Jira ticket.

- [X] PR title is either of the format **SRCH-#### \<description\>** matching the associated Jira ticket (i.e. "SRCH-123 implement feature X"), or **Release - SRCH-####, SRCH-####, SRCH-####** matching the Jira ticket numbers in the release.

- [X] Automated checks pass. If Code Climate checks do not pass, explain reason for failures:

- [X] You have provided testing instructions to help reviewers verify the changes made within the PR

[...]

PR #315, opened 2025-08-29
Staging Release SRCH-6199 

selfdanielj:
SRCH-6199 increase logging level to avoid irrelevant error logging (#314)


igoristic (approved):
Ship it :shipit:

PR #323, opened 2025-09-12
SRCH-6221 Duplicate Spider Jobs

selfdanielj:
## Summary
- Previously I prepended `Rerun::` to the jobs that were still running when we do a release in order to differentiate them from the normal scheduled jobs with the same name.
- But if a job was running when we did a release and then continues running past the time of its normal schedule, the scheduler attempts to add the job to the schedule but does not recognize that an instance is already running because of the name change.
- Instead, we can stop renaming and use native apscheduler functionality to simply update the next run time of the actual scheduled job.  It still stays on the schedule but also starts immediately.

### Testing
- Create a json file for testing that contains two or more domains on the same schedule.  Place it in `search_gov_crawler/domains`
- <details>

  <summary>Example JSON File</summary>

  ```
  [
   {
      "allow_query_string": false,
      "allowed_domains": "fac.gov",
      "check_sitemap_hours": null,
      "deny_paths": null,
      "depth_limit": 2,
      "handle_javascript": false,
      "name": "FAC.gov 1",
      "output_target": "elasticsearch",
      "schedule": "*/1 * * * *",
      "sitemap_urls": null,
      "starting_urls": "https://www.fac.gov"
   },
   {
      "allow_query_string": false,
      "allowed_domains": "gsa.gov",
      "check_sitemap_hours": null,
      "deny_paths": null,
      "depth_limit": 2,
      "handle_javascript": false,
      "name": "GSA.gov",
      "output_target": "elasticsearch",
      "schedule": "*/1 * * * *",
      "sitemap_urls": null,
      "starting_urls": "https://www.gsa.gov"
    }   
  ]
  ```
  </details>
- Set `SPIDER_SCRAPY_MAX_WORKERS=1`
- Set `SPIDER_CRAWL_SITES_FILE_NAME` to the name of your JSON file.
- Start scrapy scheduler `python search_gov_crawler/scrapy_scheduler.py`
- Confirm two jobs scheduled.
- One should start at the top of the next minute.
- Kill the scheduler and then the running job using
  - First `pgrep -f "scrapy_scheduler" | xargs kill -9`
  - Then `pgrep -f "scrapy crawl" | xargs kill -9`
- Run `python ../scripts/cache_tools.py show-jobs --pending` to show jobs left in the pending queue after shutdown, should be two jobs
- Start scrapy scheduler again
  - Whatever job was running should start right away (not wait for the top of the minute)
  - When it gets to the top of the minute you should see messages like "Execution of job X skipped: maximum number of running jobs reached"
- Kill scheduler and running jobs like above
- Recheck the pending jobs... there should still be only two

### Checklist

#### Functionality Checks

- [X] You have provided testing instructions to help reviewers verify the changes made within the PR

#### Process Checks

### DIFF
--- a/tests/scheduling/test_jobstores.py
+++ b/tests/scheduling/test_jobstores.py
@@ -40,8 +40,8 @@ GET_ALL_PENDING_JOB_TEST_CASES = [
         None,
         (
             "Found and retrieved 0 pending jobs from key test_pending_jobs",
-            "Job job1 not found in job store",
-            "Job job2 not found in job store",
+            "Job test::job1 not found in job store",
+            "Job test::job2 not found in job store",
         ),
     ),
 ]
@@ -67,6 +67,6 @@ def test_get_all_pending_jobs(
     monkeypatch.setattr(mock_redis_jobstore.redis, "zrange", lambda *_args, **_kwargs: pending_job_ids)
 
     with caplog.at_level("DEBUG"):
-        mock_redis_jobstore.get_all_pending_jobs(rerun_prefix="test::")
+        mock_redis_jobstore.get_all_pending_jobs()
 
     assert all(expected_message in caplog.messages for expected_message in expected_messages)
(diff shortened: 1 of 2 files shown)


igoristic (approved):
Nice catch. Looks good 🏆

PR #435, opened 2026-04-23
Srch 6338 ubuntu upgrade azhar

amian-gsa:
## Summary
- This PR updates the deployment scripts to make them compliant to Ubuntu 24.04 conventions and directory structures.

### Testing
- Healthcheck of instance and successful pipeline run along with other tests.

### Checklist

#### Functionality Checks

- [ ] You have provided testing instructions to help reviewers verify the changes made within the PR

#### Process Checks


cicd-scripts/app_install.sh
```diff
@@ -42,64 +82,71 @@ install_python() {
...
 # Check and install Python if needed
 check_python() {
-    if ! command -v python${SPIDER_PYTHON_VERSION} &>/dev/null; then
-        install_python
-    else
-        echo "Python ${SPIDER_PYTHON_VERSION} already installed: $(python${SPIDER_PYTHON_VERSION} --version)"
-    fi
+    log_info "Python version: $(python3 --version)"
+    # Ubuntu 24.04 has python3.12, just install venv packages
+    sudo apt-get install -y python3-venv python3-dev python3-pip
```
selfdanielj: keeping our python version separate allows us to upgrade independent of the system version.  can we keep the existing install step?

amian-gsa: It's only installing venv packages and using the python3 version that comes with the OS. This was one of the errors I was seeing where .venv directory was not found where expected. This ensure that .venv directory is present.

cicd-scripts/app_stop.sh
```diff
@@ -7,28 +7,42 @@ source ./cicd-scripts/helpers/ensure_executable.sh
...
+        echo "Error: $script_path not found!"
+        return 1
+    fi
+
+    sudo chmod +x "$script_path"
+    sudo chown -R "$(whoami)" "$script_path"
+    echo "$script_path is now executable."
+    bash "$script_path"
+}
+
```
selfdanielj: isn't this is the same as `ensure_executable` in `cicd-scripts/helpers/ensure_executable.sh`?  I don't think we need both.  if you like this formatting or name better maybe move it into the helper file?

amian-gsa: I don't see `cicd-scripts/helpers/ensure_executable.sh` in this repo.

selfdanielj: https://github.com/GSA-TTS/searchgov-spider/blob/SRCH-6338-ubuntu-upgrade-azhar/cicd-scripts/helpers/ensure_executable.sh

cicd-scripts/app_install.sh
selfdanielj: yes, I'm saying that i don't think we should use the version that comes with the OS.  It makes it much harder to upgrade later.  We create and populate that directory in app_install.sh so not sure why using the OS python would help with that.

selfdanielj (approved):
Looks good!

amian-gsa:
Changes will be back ported from staging to main after comparison and cherry picking. Closing this for now.

PR #464, opened 2026-06-15
SRCH-6348 bump pypdf

selfdanielj:
## Summary
- Update pypdf to latest
- Resolves #445 

### Testing
- Based on previous pypdf update PRs such as #390, run a PDF heavy domain, watch for errors, verify document content is loaded into opensearch:
- Ensure search services are running (at least redis and opensearch).
- Rebuild spider docker container to ensure new library updates:
   ```
   make docker-build
   ```
- Kick off pdf heavy crawl:
   ```
   make docker-scrapy-crawl ARGS="domain_spider -a allowed_domains=data.boem.gov -a start_urls=https://www.data.boem.gov/Other/FileRequestSystem/ScanPlans.aspx -a output_target=opensearch -a depth_limit=2"
   ``` 
- Watch log output for any errors and/or evidence of documents being written to opensearch (e.g. `Loaded 50 records to Opensearch!"`). This may take a while due to size of PDFs.  Its not necessary to let this crawl run until its finished.
- Check the most recent  PDFs from data.boem.gov in opensearch to see that they contain content
   ```
   GET /development-i14y-documents-searchgov/_search
   {
     "query": {
       "bool": {
         "must": [
           {
             "term": {
               "mime_type": {
                 "value": "application/pdf"
               }
             }
           },
           {
             "term": {
               "domain_name": {
                 "value": "data.boem.gov"
               }
             }
           }
         ]
       }
     },
     "sort": [
       {
         "updated_at": {
           "order": "desc"
         }
       }
     ]
   }
   ``` 

### Checklist
Please ensure you have addressed all concerns below before marking a PR "ready for review" or before requesting a re-review. If you cannot complete an item below, replace the checkbox with the ⚠️ `:warning:` emoji and explain why the step was not completed.

#### Functionality Checks

- [x] You have merged the latest changes from the target branch (usually `main`) into your branch.

- [x] Your primary commit message is of the format **SRCH-#### \<description\>** matching the associated Jira ticket.

- [x] PR title is either of the format **SRCH-#### \<description\>** matching the associated Jira ticket (i.e. "SRCH-123 implement feature X"), or **Release - SRCH-####, SRCH-####, SRCH-####** matching the Jira ticket numbers in the release.

- [x] Automated checks pass. If Code Climate checks do not pass, explain reason for failures:

- [x] You have provided testing instructions to help reviewers verify the changes made within the PR

#### Process Checks

- [x] You have specified at least one "Reviewer".

PR #472, opened 2026-06-23
SRCH-6565 freshness spider memory

selfdanielj:
## Summary
- The goal of this PR is to reduce the memory consumption of the freshness spider by letting it fetch matching documents from opensearch in batches when needed instead of all at once.
- The main change is in `search_gov_crawler/search_gov_spiders/spiders/freshness_spider.py` where the switch is made from all docs at once to batches.
- Additional Changes:
  - Removed de-duplication tracking because the freshness source docs can already generally assumed to be free of duplicates.  This lowers memory requirements.
  - Spit the pipeline file into two: one for domain spiders and another for freshness spider.  The NLTK imports in the domain spider are heavy and are not needed for the freshness spider.
  - Upgraded scrapy to latest to take advantage of some potential memory-related improvements and to address some security alerts.
  - Un-commented out the recent start and stop script changes so that the freshness spider runs again.

### Testing
- To verify a lower memory footprint I used the `memray` library and produced these before and after charts.  As you can see overall resident memory usage is down from over 350MB to under 200MB for the same set of 10k documents.  While this is not that big of a difference, this test was limited to 10k documents due to how long it takes to run a larger set.  The problem we saw was when a very large set of documents was requested to be checked for freshness.  With this change, no matter the number of documents, we should see a fairly constant memory usage.

**BEFORE**
image

**AFTER**
image

- To run your own test of the freshness spider you should:
  - `make docker-scrapy-crawl ARGS="freshness_spider -a query='{ \"query\": { \"match_all\": {} } }' -a max_results=100"`
  - And then check the `spider-freshness` index for results.

### Checklist

#### Functionality Checks

- [X] You have provided testing instructions to help reviewers verify the changes made within the PR

#### Process Checks


cicd-scripts/app_start.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -euo pipefail
+set -eo pipefail
```
luisgmetzger, line 3: Any reason for dropping `-u` here (was `set -euo`)? Guessing re-enabling the freshness start line surfaced an unset var — if so, might be safer to guard just that one (`${VAR:-}`) rather than lose unset-var detection for the whole script. Not blocking, just want to make sure it was intentional.

Makefile
```diff
@@ -103,3 +103,8 @@ docker-run-sitemap-monitor:
 		docker run -it --name $(SITEMAP_MONITOR_CONTAINER) --env-file $(ENV_FILE) --network host \
 			--volume $(VOLUME_PROJECT_ROOT) -w $(CRAWLER_DIR) $(IMAGE_NAME) python run_sitemap_monitor.py; \
 	fi
+
+.PHONY: docker-scrapy-crawl
+docker-scrapy-crawl:
```
luisgmetzger, line 108: Heads up — this exact `docker-scrapy-crawl` target also lands in #464 and #480, so two of the three will conflict on merge. Probably worth keeping it in whichever merges first and rebasing the others.

luisgmetzger (approved):
Really clean take on the memory issue — moving to the `spider_idle` + batched-injection pattern (250 at a time) instead of loading everything into the scheduler makes a lot of sense, and the memray before/after (350MB -> <200MB) backs it up. Splitting the freshness pipeline out from the NLTK-heavy domain pipeline is a nice touch too, and the new batch paths are well covered. CI's green. Approving. A couple inline notes below, plus two tiny non-blocking nits: the `""" """` empty docstring on `crawl_next_batch`, and the leftover `# with contextlib.suppress(RequestError):` comment that got carried into freshness_pipeline.py.

